PR #5, opened 2017-10-24
Update and rename createlayers.md to createlayers_zh-cn.md

yourhe:
Translated into Simplified Chinese


i0natan:
@yourhe thanks for helping on translating this!

 I believe you renamed the English file instead of creating a new translated file. Kindly create a new translated file and I'll link to it

i0natan:
@yourhe would love to work with you on the translation. Any idea how this should look like? links at the top for different languages? duplicate files for each language?

i0natan:
@yourhe since there was no response for long time I'll close the PR. Please open if it becomes relevant again

PR #6, opened 2017-10-30
Fix incomplete function declaration for appError

kevinrambaud:
Hey, I just noticed an incomplete function declaration 😁


i0natan:
@kevinrambaud good catch! :)

PR #8, opened 2017-11-02
Fix ambiguous typo in 4.1

mfine15:
I believe a verb like "using" was accidentally dropped from 4.1, making the Postman description hard to parse.


i0natan:
@mfine15 Thanks!

PR #20, opened 2017-11-04
Fixed some typos

BrunoScheufler:
Fixed some typos in the main README file

PR #21, opened 2017-11-04
fix(readme): typos

t1st3:
echo-system => ecosystem
vulenerable => vulnerable


t1st3:
fix `echo-system` typo is already suggested in #20

i0natan:
@t1st3 thanks!

PR #23, opened 2017-11-04
[errorhandling] Format js code in the blog section

kudapara:
* Format js code
* Quote the remaining paragaph from the blog post
* Remove unecessary text 'Close GIST window Skip to toolbar About WordPress'


i0natan:
@kudapara thanks a lot

PR #27, opened 2017-11-06
Fix typo section 4.7

matchai:


i0natan:
@matchai thank for, good catch!

PR #40, opened 2017-11-08
Removed parenthesis from SomeClassExample

dan-gamble:
Parenthesis throw a syntax error when used like this with `class`


i0natan:
@dan-gamble good catch, thanks you!

PR #48, opened 2017-11-09
Fix link formatting

reod:


idori:
Thanks for the fix @reod!

PR #60, opened 2017-11-11
Add section on healthchecks

Nevon:
This piece of advice isn't exactly node specific, but since Node is often deployed in a microservice environment, I think it's absolutely crucial to expose system health to all dependent systems. Even if you're not in a microservice environment, being able to view system health at a glance can be a tremendous relief.

I work in a company with a fleet of 100+ services across thousands of instances. We have developed a schema that all our services have to adhere to. Part of it is to expose a healthcheck endpoint that tells you the overall health of the system, as well as information about each individual check - including dependencies. This allows us to create a graph of connected services so that if there's an outage in one system, we can see which other systems will be affected by it. It can also help us discover bottlenecks where one system may affect many others, in which case we should strive to provide fallback behaviors that allow us to at least partially function.

For the record, I'm the developer of [express-physical](https://www.npmjs.com/package/express-physical), which I linked to in the "read more" section. We use that library across all of our node services, and it has worked well for us so far.

### DIFF
--- a/README.md
+++ b/README.md
@@ -650,6 +650,19 @@ All statements above will return false if used with `===`
 
 **Otherwise:** Long deployments -> production down time & human-related error -> team unconfident and in making deployment -> less deployments and features
 
+<br/><br/>
+
+
+## ![✔] 5.17. Expose a healthcheck endpoint for dependent systems
+
+**TL;DR:** While a simple ping endpoint can tell you if an application is running, it can be more difficult if it is able to handle requests. All applications in a distributed system should expose a healthcheck endpoint that returns the overall health of the system. Such an endpoint can perform a number of different checks, such as database access, available disk space and more. A monitoring service can then periodically query that endpoint and build an overall view of the health of the system.
+
+**Otherwise:** When a distributed system grows, it can become difficult to get an accurate picture of the overall system health.
+
+
+🔗 [**Read More: Expose a healthcheck endpoint for dependent systems**](/sections/production/healthcheckendpoint.md)
+
+
 <br/><br/><br/>
 
 <p align="right"><a href="#table-of-contents">⬆ Return to top</a></p>
--- a/sections/production/healthcheckendpoint.md
+++ b/sections/production/healthcheckendpoint.md
@@ -1,0 +1,12 @@
+# Expose a healthcheck endpoint for dependent systems
+
+<br/><br/>
+
+
+### One Paragraph Explainer
+
+In a distributed system, transparency regarding instance health is of utmost importance. If a customer action fails because one system in a chain of multiple is unhealthy, it can be very difficult to know where the problem is.
+
+By exposing an endpoint that gives you a picture of the health of the system, a system monitoring application can query all involved systems to find out which ones are unhealthy and what dependent systems may be affected by an outage.
+
+[Adopt a standard schema](https://www.npmjs.com/package/express-physical) for these endpoints across your applications to easily be able to monitor the system as a whole. 


i0natan:
@Nevon interesting point, great to have you here

We already have 'Create maintenance endpoint' (5.7), doesn't it overlap? is it right that when DB is down all microservices will report 'unhealthy' state? what else can healthpoint check besides disc-space?

Nevon:
There is overlap between a maintenance endpoint (I would call that a metrics endpoint), but it does not replace a healthcheck. If metrics are data, then a healthcheck is information - the interpretation of data. It also allows you to raise your view from just an instance to a system, as it allows you to track problems across components of a system.

> is it right that when DB is down all microservices will report 'unhealthy' state?

Depends how the system is designed, but in most scenarios that would mean that one service would be down, which may or may not affect dependent systems. But let's say that the authentication service in your system is failing to reach its database. Suddenly customers are calling in complaining that they are unable to log in to your product. At this point someone would contact the product team, which would investigate their logs and see that the authentication service is responding with a 500, after which they will contact the authentication service team, which would investigate and see that they are unable to reach their database.

If all services had exposed healthchecks with explicit dependencies, your monitoring system could have alerted both teams, and it would have been immediately visible that the error was not actually in the product service, and that they were just affected by an outage in the authentication service.

> what else can healthpoint check besides disc-space?

Whatever is relevant to your service. In my solution, I broadly categorize them into 6 categories:

* SELF - problem with the application itself, i.e. crashed
* METRICS - could be business level events, i.e. no orders placed
* INFRASTRUCTURE - DB, local process, work queue, etc.
* INTERNAL_DEPENDENCY - "internal" meaning within the company, i.e. the authentication service doesn't respond
* EXTERNAL_DEPENDENCY - third party dependency, like an external API not responding
* INTERNET_CONNECTIVITY - what it says on the tin

This is by no means a canonical list in any way, just a useful categorization. Each application will have its own needs.

As an example, one service I work on has the following health checks (from memory, so I'm sure I'm missing stuff):

* Redis connectivity
* Database connectivity
* Disk < 80% full
* Available heap space
* Remote configuration can be read
* Can connect to a local agent of another service
* Can write to log file
* Can connect to Kafka
* Available database connections

i0natan:
@Nevon sorry for the late reply, missed your insightful answer.

I have two questions:
1. How can this plug into the standard monitoring platforms - you mentioned the need for system-wide view of information, achieving that require integration with some popular platform like ELK, Cloudwatch, etc. These platforms won't approach this API endpoint rather demand receiving logs/events. Wouldn't it be right to emit this info constantly, along with other info such as hardware status (using elk heartbeat for example) and consume everything from a single dashboard (e.g. Kibana)?

2. If services A, B & C are using PostgreSQL and it's currently down, what should happen when their report this information? if they return fauly state to the ELB all will get down

Nevon:
1. It's kind of an inversion of control, and in my mind the two approaches can and should coexist. If you push your logs to something like ELK, the knowledge and responsibility of determining if an application is healthy is moved from the application itself to the external system. A healthcheck endpoint allows the application itself to self-determine if it is healthy, because most likely it is the one that is best able to tell whether not it is. There's also the consideration of what happens if the monitoring platform is down. If the healthcheck is done within the application, then at the very least you can still monitor it with cURL, so you're not completely blind.

However, what I advocate and practice is to do both. The application can determine if it thinks it is healthy, and then an external log processing platform can also give a picture of system health by observing system level metrics and business events (for example, if the ratio between created orders and completed orders changes drastically, then something may be wrong somewhere).

2. Personally I don't use this endpoint to report healthiness to the ELB. I just use it to alert responsible teams that they need to take an action. But if you have fine-grained enough metadata regarding your healthchecks (some are critical, some are not), you could potentially use it from the ELB. The problem is that only a few classes of problems will be fixed by replacing the instance, which is why I'm not using it like that.

But honestly, a lot of this is application dependent. I don't think there's a perfect solution that works the same for all cases.

i0natan:
1. You should separate instance and application. Health checks occur at the instance level, meant for load balancing purposes only, imply real-time actions and their outcome might be only - instance is being replaced. Application-wide monitoring spans all the pieces, meant for any kind action and insight and usually leaves the outcome for the ops team. If you agree with that- then healthcheck should only collect metrics that might take the instance down like disc-space. Any other metrics that are related to other services, like DB connection, should take place somewhere else and be written to logs so the ops can act on it. Systems like ELK contain many feaures for acting on data like summarizing the amount of errors per minute and sending alerts to Slack
2. Agree. If service A can't approach service B (=business microservice, DB, mq, whatever), then service A shouldn't mark itself as unhealthy rather report a circuit-break, try to be resilience and operate the best it can. Otherwise will have a domino effect...

PR #61, opened 2017-11-12
Remove unneeded image and replace with code block

hisaac:
I noticed there was an image of some code that would be nicer as an actual code block, so I just transcribed what was in the image, and deleted the image from the repo.


refack:
@hisaac thank you for the contribution.

hisaac:
You're welcome! Glad to clean up little things when I see them. I'm slowly making my way through the articles here, and they've been really helpful.

Issue #50, opened 2017-11-10
should probably suggest `rc` for 1.5


nwwells:
[`rc`](https://www.npmjs.com/package/rc) is a more popular config tool than `nconf` and `config`, is used by many rock-solid tools out there, and checks all the boxes you require. I'd suggest recommending it for config management.

i0natan:
@nwwells good point!

can you elaborate how it stacks against those two? any differences/downside?

nwwells:
I've never used the other two, tbh, but it supports everything I saw in the brief look that I had at the other two, including:

file (with recursive up-tree search)
option parsing
env vars
programmatic defaults and overrides

i0natan:
it also more popular, makes sense to add it, can you kindly PR?

nwwells:
Will do. Hopefully tonight

nwwells:
Closed with #66

PR #66, opened 2017-11-14
add rc as config manager option

nwwells:
See #50


refack:
@nwwells Thanks for following up 🥇

PR #75, opened 2017-11-15
Fixed some typos

paulovitin:
Fixed some text and code example typos.


BrunoScheufler:
I updated some syntax too! Thanks for finding those issue 👍

PR #79, opened 2017-11-20
The == result does not change when args swap

caribe:


idori:
Hi @caribe, thanks for the input!
You are correct, but please note that is line and the next are not just swapping the arguments , first has '0' as a string and the next one as a number

`'' == '0'` does indeed return `false`

idori:
closed

PR #81, opened 2017-11-23
Update README.md

### LINKED ISSUE #72: 3.6 Naming conventions for variables, constants, functions and classes
```js
// for const name we use UPPERCASE
    const CONFIG = {
      key: 'value'
    };
```
I would like to mention that this is legacy since first versions of fortran.
It didn't have const's! 
So for all registries/variables that we didn't wanna ppl to change we started using UPPERCASE notation.
Again: it was just because language didn't support `const`!

It's not `nice` to see that this legacy lives on 30+ years, even when most major languages/compilers understand `const` word. 

Also most editors understand word `const` and display it different than variables (let, var).

I would postulate to drop 30+ years legacy, and rethink how we are doing things, and change:
```js
// for const name we use UPPERCASE
const CONFIG = {
      key: 'value'
};
```
to
```js
const config = {
   key: 'value'
}
```

syzer:
Fixes #72

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,13 +286,13 @@
   // for class name we use UpperCamelCase
   class SomeClassExample { 
     
-    // for const name we use UPPERCASE
-    const CONFIG = {
-      key: 'value'
+    // for const name we use const keyword
+    const config = {
+      key: `value`
     };
     
     // for variables and functions names we use lowerCamelCase
-    let someVariableExample = 'value';
+    let someVariableExample = `value`;
     function doSomething() {
       
     }


README.md
```diff
@@ -286,13 +286,13 @@
   // for class name we use UpperCamelCase
   class SomeClassExample { 
     
-    // for const name we use UPPERCASE
-    const CONFIG = {
-      key: 'value'
+    // for const name we use const keyword
```
refack: names

syzer: ?

refack:
Hello @syzer and thank you for the contribution 🥇 
Maybe we should consider merging those two paragraphs (all are variables, and named in the same style)?

README.md
refack: `for const names we use the const keyword`
(either `a const name` or `const names`)

syzer: https://github.com/i0natan/nodebestpractices/pull/82

i0natan:
@refack I'm not sure whether we fixed the permission issue - can you know merge PR?

@syzer - thanks for this nice contribution :)

syzer:
Closing in favour of #82

PR #82, opened 2017-11-23
Update README.md

syzer:


BrunoScheufler:
Thanks for updating this! By the way, you could also have added this to the other PR as a single commit to the same branch but no worries there 👍

PR #85, opened 2017-11-25
Add Prettier to Code Style Practices

septa97:


i0natan:
@septa97 thanks for joining the battle :)

I'd like to clarify that a bit - why prefer prettier over eslint fixes? they at least overlap, no? should we mention vendors like this or maybe focus on the need and the less explicitly at the bottom specify few vendors?

septa97:
Prettier is made specifically to format code unlike ESLint that format codes and at the same time, it checks for possible code errors (linting). My practice in my projects is to use Prettier to format the code and to use ESLint to check for errors.

[Here](https://github.com/prettier/prettier-eslint/issues/101) is the link of an instance where Prettier performs better than ESLint in code formatting.

And yes, Prettier and ESLint overlaps in the code formatting feature but it can be easily solved by using other packages like [prettier-eslint](https://github.com/prettier/prettier-eslint), [eslint-plugin-prettier](https://github.com/prettier/eslint-plugin-prettier), and [eslint-config-prettier](https://github.com/prettier/eslint-config-prettier). For more information about their difference, you can view the link [here](https://stackoverflow.com/questions/44690308/whats-the-difference-between-prettier-eslint-eslint-plugin-prettier-and-eslint).

septa97:
Those 3 packages that I mentioned are for Prettier and ESLint integration.

septa97:
Just a question, do you think I should just include this in section 3.1 together with ESLint? Since they are meant to be used together.

i0natan:
@septa97 sorry for the late reply, somehow missed this useful thread.

I think the case for prettier is clear and valuable, I would include it as part of ESLint:
- Within the TL&DR would mention briefly that it can auto-fix automatically and consider using Prettier which speciailizes in fix
- Within the details page elaborate more, include code example (the one you provided is great) and link to the extensions you provided

cool?

thanks!

septa97:
Done changing the PR. I'll kindly wait for your review and comments.

i0natan:
@septa97 I merged the PR, great contribution.

Had to perform two changes because both are not aligned with our writing guidelines which we should publics in few days... you couldn't know, this is why I explain here:
1. I left only ESLint in the title - we recommend a specific and single tool when it absolutely 5X time mothe re popular than others. Prettier seems great but has competitors and it doesn't control this segment like ESLint does. It worth a recommendation but at the same level with ESlint
2. The detail page wasn't done with our template, it has different sections - all pages should have similar structure. I'll send you the template today and let's format this page then link to it?

Anyway, great addition, you've your flower at the home page :)

PR #96, opened 2017-12-04
[Chinese translation]Readme Chinese translation part one

MattJin:


i0natan:
@MattJin don't you have permissions to merge into the chinese branch?

PR #102, opened 2017-12-06
Update frontendout.md

Sargon2234:


BrunoScheufler:
Thanks for the contribution, would you like your name on the 'Thank You Notes' list at the bottom of the main README? You can just append your name with the same scheme as the other names and push to this same branch if you want 👍

PR #117, opened 2017-12-27
Optimized sytax and fixed typos

BrunoScheufler:


BrunoScheufler:
@i0natan thanks for looking over it 😄!

PR #118, opened 2017-12-29
Fix wrong copypasta

Kiruchi:
This code is not Nginx but Express


BrunoScheufler:
👍 You're actually right, I didn't see this one in my changes. Thanks for that!

Kiruchi:
Thank you for asking, I will do that if I do a major change, but not on this one ;)

PR #144, opened 2018-02-17
Expand 6.1 to include database libraries

chrisnicola:
Added details about parameter binding support in typical database libraries, as ORM/ODM libraries are not the only options to protect against injections. Also included a sentence about explicitly avoiding template strings or string concatenation to create queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -656,8 +656,9 @@ All statements above will return false if used with `===`
 
 # `6. Security Best Practices`
 
-## ![✔] 6.1. Use an ORM/ODM to prevent SQL/noSQL injection attacks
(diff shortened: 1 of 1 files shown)


i0natan:
@chrisnicola welcome aboard! Technically the addition absolutely makes sense. You don't have to use ORM to be secured.

I would just make the texts a bit more fun & easy to read. We have too many if/and + the title is too lengthy. Any chance you rephrase it? we can drop some details, there will be 'read more' page

p.s. going by our procedure, I assigned you a flower on our homepage :)

chrisnicola:
🥂

PR #158, opened 2018-03-18
[Chinese-translation]some updates

MattJin:


i0natan:
@MattJin great!

i0natan:
@MattJin shall we publish this in few Chinese JS/Node/Mean groups?

MattJin:
@i0natan I think it's ok.

PR #161, opened 2018-03-25
MINOR 3x security bullets

js-kyle:


i0natan:
@js-kyle great stuff there, one remark about

"Support JWT tokens blacklist (passport won't provide that by default)"

I intentionally added the passport context to avoid writing a generic/non-node title, you may change the text but try to provide some node context into the generic security advice. makes sense?

js-kyle:
@i0natan Ah, yes - makes sense. Will add some Node.js context into that bullet

PR #166, opened 2018-03-31
Security 6.8 Avoid using the Node.js Crypto library for passwords

js-kyle:


sections/security/bcryptpasswords.md
```diff
@@ -0,0 +1,28 @@
+# Avoid using the Node.js Crypto library for passwords, use Bcrypt
+
+### One Paragraph Explainer
+When storing user passwords, using an adaptive hashing algorithm such as bcrypt, offered by the [bcrypt npm module](https://www.npmjs.com/package/bcrypt) is recommended as opposed to using the native Node.js Crypto module. `Math.random()` should also never be used for as part of any password or token generation due to it's predictability.
+
```
i0natan: @js-kyle I believe that many readers won't easily grasp the meaning of 'number of rounds can be specified'

Maybe a short explanation, highly friendly and abstracted, can make the explanation clearer to readers who are not familiar with bcrypt idea?

js-kyle: Yep, agree with you there. I'll have a think about the best way to do this

i0natan: Maybe explain that using standard crypto algorithms an attacker can try many combinations very quickly, Bcrypt was born to address this and it's slow by design to limit the number of attempts/sec

sections/security/bcryptpasswords.md
```diff
@@ -0,0 +1,28 @@
...
+The `bcrypt` module or similar should be used as opposed to the JavaScript implementation, as when using `bcrypt`, a number of 'rounds' can be specified in order to provide a secure hash. This sets the work factor or the number of 'rounds' the data is processed for, and more hashing rounds leads to more secure hash (although this at the cost of CPU time). The introduction of hashing rounds means that the brute force factor is significantly reduced, as password crackers are slowed down increasing the time required to generate one attempt.
+
+### Code example
```
BrunoScheufler: @i0natan @js-kyle maybe we should create and enforce a guideline for JavaScript code examples, because we'll end up with many different types of code snippets, using different ECMAScript features and so on.

i0natan: @BrunoScheufler what for example might confuse the user? I need an example of confusing examples or any existing examples that are not consistent

BrunoScheufler: @i0natan it's not so much about confusion, rather inconsistency, for example [this function](https://github.com/i0natan/nodebestpractices/pull/166/files#diff-ccb88acb49767f79fac21b4425a5e551R17) uses a function literal, whereas we also use arrow functions in other cases. I think it's about the things that might look a bit nitpicky but make a difference viewed as a whole.

i0natan: @BrunoScheufler see your point, I'm just no sure how "structured and unified" we wanna be. Where would you maintain these guidelines? won't this make our writing guidelines too long and maybe, as you said, nitty gritty? I absolutely agree that it's better to have the code examples unified, just not sure whether we should be so strict

PR #171, opened 2018-04-09
fix bare markdown

keyfoxth:


BrunoScheufler:
Thank you very much @keyfoxth!

i0natan:
@keyfoxth thanks, flower will be assigned on April end

#172

PR #174, opened 2018-04-18
Security - run as non root

i0natan:
@lirantal @js-kyle @BrunoScheufler


BrunoScheufler:
Will have a look at it in a few minutes!

README.md
```diff
@@ -786,12 +786,12 @@ Developers in the project may not follow consistent code security practices, lea
 
 ## ![✔] 6.13. Run Node as non-root user
 
-**TL;DR:** Running on behalf of the root or full-privilege account means that any attacker can gain full control to perform anything within the machine if your application is compromised.
+**TL;DR:** In Docker environment, Nodejs runs as a root user by default with unlimited permission set. Create a noon-root user during image creation and run the container with the flag "-u username"
```
lirantal: @i0natan I'm not sure if I'd specifically refer to Docker here since many people just run the Node.js process on a barebones machine (i.e: no virtualization).

With that said, you're more detailed explanation is pretty good and provides a general description of the problem so either leave as is or just rephrase a bit.

sections/security/non-root-user.md
```diff
@@ -0,0 +1,30 @@
+# Run Node as Noon-Root User
+
+### One Paragraph Explainer
+According to the 'Principle of least privilege' a user/process must be able to access only the information and resources that are necessary. Granting root access to an attacker opens a whole new world of malicious ideas like routing traffic to other servers. In practice, most Node apps don't need a root access and don't run with such privilege. However, there are two common scenario that might push to root usage: (1) to gain access to privilege port (e.g. port 80) Nodejs must run as root (2) Docker containers by default run as root(!). It's recommended for Node webserver to listen to non-privilege ports and rely on a reverse-proxy like nginx redirect from port 80. When building a docker image, highly secured apps should run the container with an alternate non-root user. Most docker clusters (e.g. Swarm, Kubernetes) allow setting the security context declaratively
```
lirantal: - s/user/process
- s/node/Node.js
- s/nodejs/Node.js

sections/security/non-root-user.md
```diff
@@ -0,0 +1,30 @@
+# Run Node as Noon-Root User
+
+### One Paragraph Explainer
+According to the 'Principle of least privilege' a user/process must be able to access only the information and resources that are necessary. Granting root access to an attacker opens a whole new world of malicious ideas like routing traffic to other servers. In practice, most Node apps don't need a root access and don't run with such privilege. However, there are two common scenario that might push to root usage: (1) to gain access to privilege port (e.g. port 80) Nodejs must run as root (2) Docker containers by default run as root(!). It's recommended for Node webserver to listen to non-privilege ports and rely on a reverse-proxy like nginx redirect from port 80. When building a docker image, highly secured apps should run the container with an alternate non-root user. Most docker clusters (e.g. Swarm, Kubernetes) allow setting the security context declaratively 
+
+### Code example - Building a Docker image as noon-root
+```javascript
+FROM node:some-version-here
```
lirantal: I'd just specify a version instead of the placeholder to make it more readable

BrunoScheufler: I'd agree to @lirantal, even the 'latest' tag would suffice for this, the rest seems pretty good though!

js-kyle:
Nice work @i0natan

sections/security/non-root-user.md
i0natan: @lirantal @BrunoScheufler good point

README.md
i0natan: @lirantal I'm splitted here since 95% who runs node in bare metal doesn't run them as root, while many people use Docker and 95% of them run node as root...

sections/security/non-root-user.md
i0natan: @lirantal probably missed something, to what these paths refer to?

README.md
lirantal: Like in my original comment, I'm also split between mentioning or not :-)
Thing is, you are being very specific. For example, what if technology changes tomorrow and docker is gone in favor of kube or something else? That's why I don't like this buzz word name dropping thing. Maybe replace "docker" with "container technology" or just "virtualization".

PR #178, opened 2018-05-06
6.3 Extract secrets from config files or use NPM package that encrypt…

js-kyle:
…s them

### DIFF
--- a/README.md
+++ b/README.md
@@ -772,14 +772,13 @@ Developers in the project may not follow consistent code security practices, lea
 <br/><br/>
 
 
-## ![✔] 6.12. Avoid using middlewares that crash the process
+## ![✔] 6.12. Implement express rate limiting for login routes
 
-**TL;DR:** Here we will write about designing and using middleware that doesn't make the process crash when an invalid JSON is passed since it opens the door for easy DDOS attacks
+**TL;DR:** A brute force protection middleware such as [express-brute](https://www.npmjs.com/package/express-brute) should be used inside an express application to prevent brute force/dictionary attacks on sensitive routes such as `/admin` or `/login` based on request properties such as IP address, or other identifiers such as body parameters.
 
-**Otherwise:**
+**Otherwise:** An attacker can issue unlimited automated password attempts to gain access to privileged accounts on an application.
 
-
-🔗 [**Read More:**](#)
+🔗 [**Read More: Login rate limiting **](/sections/security/login-rate-limit.md)
 
 <br/><br/>
 
--- a/README.md
+++ b/README.md
@@ -684,7 +684,7 @@ Developers in the project may not follow consistent code security practices, lea
 **Otherwise:** Source control for even private repositories, can mistakenly be made public, at which point all secret has been exposed outside. Access to source control for an external party will inadvertently provide access to related systems (database, apis, etc).
 
 
-🔗 [**Read More:**](#)
+🔗 [**Read More: Secret management**](sections/security/secretmanagement.md)
 
 <br/><br/>
 
--- a/sections/security/secretmanagement.md
+++ b/sections/security/secretmanagement.md
@@ -1,0 +1,31 @@
+# Extract secrets from config files or use NPM package that encrypts them
+
+### One Paragraph Explainer
+The most common and secure way to provide a Node.js application access to keys and secrets is to store them using environment variables on the system where it is being run. Once set, these can be accessed from the global `process.env` object. 
+A litmus test for whether an app has all config correctly factored out of the code is whether the codebase could be made open source at any moment, without compromising any credentials.
+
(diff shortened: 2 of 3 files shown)


js-kyle:
cc @i0natan @BrunoScheufler 

Also @BrunoScheufler do we prefer npm or NPM? We use both widely across the project so I could update these all to be consistent

BrunoScheufler:
@js-kyle I'd say we should use npm because I believe it's the official capitalization of the service. I could update that later :+1:

sections/security/secretmanagement.md
```diff
@@ -0,0 +1,31 @@
+# Extract secrets from config files or use NPM package that encrypts them
+
+### One Paragraph Explainer
+The most common and secure way to provide a Node.js application access to keys and secrets is to store them using environment variables on the system where it is being run. Once set, these can be accessed from the global `process.env` object. 
+A litmus test for whether an app has all config correctly factored out of the code is whether the codebase could be made open source at any moment, without compromising any credentials.
```
i0natan: Great explanation (the litmus thing)

sections/security/secretmanagement.md
```diff
@@ -0,0 +1,31 @@
...
+
+For rare situations where secrets do need to be stored inside source control, using a package such as [cryptr](https://www.npmjs.com/package/cryptr) allows these to be stored in an encrypted form as opposed to in plain text.
```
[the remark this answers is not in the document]
i0natan: @alonbardavid Welcome!

We all agree (do we?) that using a vault is the prominent option. Alternatives can be considered to small projects, is git encryption a valid option for small project?

To my & @js-kyle understanding it's a valid option - help us realize why not: the decryption-key can be shared via secured channel (e.g. a phsical note), it can be shared only with few developers or with devops only, it's better than nothing. What do you think?

@js-kyle https://github.com/shadowhand/git-encrypt seems more popular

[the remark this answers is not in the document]
js-kyle: @alonbardavid I think I see what you're saying - that if the option is available for the decryption key to be provided at runtime , why wouldn't that method be used used to store the secret in the first place?

@i0natan I chose cryptr as I felt this article was becoming a bit too broad, git-encrypt seems less Node.js focussed, what do you think?

README.md
```diff
@@ -684,7 +684,7 @@ Developers in the project may not follow consistent code security practices, lea
 **Otherwise:** Source control for even private repositories, can mistakenly be made public, at which point all secret has been exposed outside. Access to source control for an external party will inadvertently provide access to related systems (database, apis, etc).
```
i0natan: @js-kyle We should encourage devops way of thinking - source control is for developers, the production secrets must be visible to the people who operate the production only. Even if made private, a source control is not the appropriate system to store highly sensitive data

js-kyle:
@BrunoScheufler Happy to update NPM => npm, should we make that change in master or this branch for now?

sections/security/secretmanagement.md
```diff
@@ -0,0 +1,31 @@
+# Extract secrets from config files or use NPM package that encrypts them
+
+### One Paragraph Explainer
+The most common and secure way to provide a Node.js application access to keys and secrets is to store them using environment variables on the system where it is being run. Once set, these can be accessed from the global `process.env` object.
```
i0natan: I think you should start with the concept of vaults - this what makes the real difference. Most developers do *finally put the code in env var, only they first store them in .env files within the source control... The critical part is to fetch the secrets from the vault during deployment, where only the deployment component has access to the vault, and it inject the secret into the production env var @js-kyle

js-kyle:
@i0natan Good feedback, I'll revisit some of the point brought up here mainly around the concept of a vault 👍

BrunoScheufler:
@js-kyle I think the changes to the npm/NPM problem would currently fit best into a PR to `master`, we can encourage the unified spelling in new PRs and content after that, be it to the `security-best-practices` branch or else.

sections/security/secretmanagement.md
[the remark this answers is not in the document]
alonbardavid: @js-kyle precisely. Especially since you'd want the key to the secret to have enough entropy to not be breakable on it's own, which usually means it'll be hard to remember. If it's hard to remember, you are going to put it in a password safe anyway, might as well put the actual secret there.

sections/security/secretmanagement.md
```diff
@@ -0,0 +1,31 @@
...
+There are a variety of tools available which use git commit to audit commits and commit messages for accidental additions of secrets, such as [git-secrets](https://github.com/awslabs/git-secrets).
+
+### Code example
+Accessing an API key stored in an environment variable:
+```javascript
+    const azure = require('azure');
+
+    const apiKey = process.env.AZURE_STORAGE_KEY;
+    const blobService = azure.createBlobService(apiKey);
+```
```
sagirk: I'd fix the indentation to be consistent with all JS blocks in README.md

PR #180, opened 2018-05-11
updated NPM => npm

js-kyle:


js-kyle:
@BrunoScheufler For what is in master, yes. We might have added some more in the security branch, so we can perhaps check that as part of a final review

BrunoScheufler:
@js-kyle yeah right, although its already good for now! Might be done as part of another PR later!

i0natan:
@js-kyle @BrunoScheufler I love your keen eye for details :)

@js-kyle I think it's about time we add you to the repo collab + maintainers list, no?

i0natan:
Adding to #183

Issue #187, opened 2018-05-17
Suggestion 5.7. maintenance endpoint is dangerous when it becomes a target of DDOS


serv:
Opening up an endpoint that generates MBs of Node dumps can be dangerous.
Once the endpoint is exposed, anyone can write a simple script that repeatedly hits on the endpoint and bring down the app server.
If this endpoint is exposed only for signed-in admin users, then this risk can be mitigated. You should mention this as a hard requirement for this suggestion.

i0natan:
@serv Welcome!

It's written that "A maintenance endpoint is a plain secured HTTP API" - maybe we should emphasize that his endpoint must be highly secured and besides typical authentication be opened to white-listed IPs only

Do you wanna PR something like this?

serv:
I'll mention that this should be an endpoint that is not available to public. And if someone who is not authorized attempts to request the endpoint, they get 403.

I think white-listed IPs are too stringent of a requirement. People might be debugging from places (at home, cafe etc) that might not be whitelisted. Also at work network, they might have variable IPs for users.

Issue #183, opened 2018-05-15
May 2018 Maintenance


sagirk:
Assignee: @BrunoScheufler

**Updates**

* Update top badges with best practices item count, last update date and Node.js version
* Ensure all translations are aligned with the English version

**Flowers**
@duartemendes
@js-kyle
@sagirk
@serv 

**Stars**

**Core Team**
@js-kyle

BrunoScheufler:
👍

i0natan:
Performed the May maintenance checklist

@duartemendes @js-kyle @sagirk @serv - You were assigned flowers for the great contrubition. Hope to work together on other improvements

@js-kyle Can you kindly add yourself to the core team as you have to include a BIO about yourself?

PR #188, opened 2018-05-30
#187: 5.7. maintenance endpoint is dangerous when it becomes a target…

serv:
… of DDOS

PR for issue #187


i0natan:
@serv Great! Added to #183 

@BrunoScheufler, shall we define that a PR approver ask the opener to add himself to the maintenance list?

BrunoScheufler:
@i0natan we actually did this in the past, that tradition might have slipped somewhere, let's do that again!

PR #200, opened 2018-06-23
Security - Escape output

i0natan:
@lirantal @BrunoScheufler @js-kyle


README.md
```diff
@@ -739,13 +739,13 @@ Developers in the project may not follow consistent code security practices, lea
 
 <br/><br/>
 
-## ![✔] 6.9. Use middleware that sanitizes input and output
+## ![✔] 6.9. Escape output
```
lirantal: The term `Encoding` is better IMO

BrunoScheufler: Wouldn't encoding rather fit for things like text encodings (utf8, hex, etc.) and so on?

lirantal: Yes, and in our context encoding is just fine and what we mean when related to security - we encode whatever input data to its proper context.

See reference for how OWASP Also uses encoding as the official description through-out the wiki: https://www.owasp.org/index.php/XSS_(Cross_Site_Scripting)_Prevention_Cheat_Sheet and https://www.owasp.org/index.php/OWASP_Java_Encoder_Project to name a few

BrunoScheufler: Okay I never heard it in the context of security but that's really interesting!

i0natan: @lirantal Ahlan!

Encoding sounds like the right technical word, but 'escaping' has this magical context that gives special meaning to this specific type of critical encoding. I'll use both if resonates with you

PR #216, opened 2018-07-22
add some notes about mongoose to error handling

vkarpov15:
Re: https://github.com/Automattic/mongoose/issues/6424, figured I'd start with adding a note about error handling in mongoose and link to Express' error handling docs.


i0natan:
@vkarpov15 Welcome aboard! Thanks for the contribution.

Two questions:
1. Why a data access layer is relevant in this regard? the entry point, Express/Koa, are relevant becauase they start new requests and any request-level error will bubble down to their catcher

2. Should we put sentry, a commercial product, within the text? 

See our [content writing manifest, bullet 6](https://github.com/i0natan/nodebestpractices/blob/master/writing-guidelines.md)

vkarpov15:
1. Sometimes it is helpful for an ODM or ORM to mutate errors or report errors separately, to do things like record the query that caused the error or clean up the error message to make it more relevant to application users.

2. That's fair, I wasn't aware you had a policy against including commercial products in the text. Just to clarify, I have no affiliation with Sentry, but I think that tracking errors in a tool like Sentry or Raygun is a best practice, and Sentry shows up as the number 2 result on Google for "error tracking" after Rollbar

i0natan:
First, re-iterating my appreciation for the PR. Im following your blog for a long time.

1. While I absolutely agree about the need for ODM/ORM to sanitize and contextualize errors, I believe that in this specific bullet - having a centralized error handler that is agnostic to all the layers and technologies - mentioning the DB layer might confuse the reader. That said, it strikes me that we do need to explain more on where the ODM/ORM fits in the error chain. I would do that using a dedicated bullet or over an existing one

2. Actually mentioning APMs in the sense of centralized error handler does make sense. Our policy just recommends not to favour one vendor so if we mention also Rollbar and Raygun that sounds fair. Do you want to perform this edit or shall I?

vkarpov15:
I took care of point (2), but I'm not sure what you mean by "dedicated bullet" in point (1), can you please clarify that?

vkarpov15:
Yeah absolutely, go for it.

i0natan:
Merged.

@vkarpov15 Read the buffer article, nice one. 

We're having thoughts of creating a starter that packs all these best practices (yet minimal and non-opinionated). It aims to be the Javs-spring/.net-EnterpriseLibrary of the node.js world. Any thoughts on this, ideas or willingness to join are welcome

PR #217, opened 2018-07-24
Used async/await syntax over chaining .then's

GabeKuslansky:


BrunoScheufler:
@i0natan I think this is part of a problem we've had for as long as this project exists. I really believe that we need uniform coding standards to solve issues like "should i use semicolons?", "what JavaScript features can I utilize?". If we discuss and persist these guidelines, we can save time with code refactoring PRs and can directly review new content using the guidelines.

i0natan:
@GabeKuslansky This is a great contribution. Code is so much cleaner now!

@BrunoScheufler You man to specify in advance code standards for new PRs and fix the historical content to some unified syntax? can you specify 3-4 code standards just to see that it's not too subjective? semicolon for example is highly subjective and I don't see why we should take a side in this debate

BrunoScheufler:
@i0natan let's move this thread to another place!

BrunoScheufler:
@i0natan let's create a list for the end of this month, I think that scales better!

i0natan:
@BrunoScheufler Can you?

BrunoScheufler:
@i0natan #227

Issue #227, opened 2018-08-16
August maintenance


BrunoScheufler:
**Flowers**
- GabeKuslansky
-  [Ron Gross](https://github.com/ripper234)
- [Valeri Karpov](https://github.com/vkarpov15)

**Stars**

js-kyle:
@BrunoScheufler Do you want me to make this update?

PR #230, opened 2018-08-17
August maintenance

js-kyle:


i0natan:
@js-kyle The maintenance list is an issue, noted that @BrunoScheufler already opened one here #227 

unless you suggest to change the flow which is always welcome

PR #232, opened 2018-08-19
MINOR Replace nsp with npm audit

js-kyle:


i0natan:
@js-kyle @BrunoScheufler Do you have permission to merge?

js-kyle:
@i0natan Yes I do, so I can just merge stuff like this if we would prefer that

i0natan:
@js-kyle @BrunoScheufler Let's discuss on Saturday but in overall seems to me like we should always have 1 approver having a look before a merge (another pair of eyes)

PR #237, opened 2018-09-04
Use ES6 async/await in the example

imsergiobernal:
Hi, I'm a beginner developer. I think now it's more clear with ES6 Async/await and arrow function.

I didn't use try-catch because throw. I'm just practising! Any suggestion or fix is very welcome


js-kyle:
Welcome @sbernaldev, Thanks for the contribution - we'll take a look at this very soon!

i0natan:
@js-kyle Can I take this (have some weird sympathy for error handling?

@sbernaldev Welcome! We were mistaken to show an example with 'reject' - throw is much better as it enforces one strict way of interrupting the program flow (reject doesn't even stop the flow). This is an opportunity to fix this. Wanna perform the change?

js-kyle:
@i0natan Yep go for it 👍

sbernaldev:
Perfect @i0natan! Thanks a lot for your advice. Something like this would be correct? I prefered to ask you before commiting changes, is this the best way? I'm beginner contributing to public repos!

`
const addProduct = async (productToAdd) => {
try {
  const existingProduct = await DAL.getProduct(productToAdd.id);
  if (existingProduct != null)
      throw new Error("Why fooling us and trying to add an existing product?"));
} catch (err) {
  ...
}
`

sagirk:
@sbernaldev Looks good! I have a few minor tweaks to suggest below. Also, the error message's language feels like it blames the user of malicious intent. I would reword to make it more neutral.

```diff
  const addProduct = async (productToAdd) => {
    try {
      const existingProduct = await DAL.getProduct(productToAdd.id);
-     if (existingProduct != null)
+     if (existingProduct !== null) { // Better or...
+     if (existingProduct) {          // ...simpler
-       throw new Error("Why fooling us and trying to add an existing product?"));
+       throw new Error("Product already exists!"));
+     }
    } catch (err) {
-   ...
+     // ...
    }
  }
```

sbernaldev:
Thanks a lot for joining @sagirk! I totally agree with your suggestions. Also I would use that "...simpler" way because it will throw that error only when _if statement_ is truthy (and not when undefined or null, for example).

```javascript
const addProduct = async (productToAdd) => {
  try {
    const existingProduct = await DAL.getProduct(productToAdd.id);
    if (existingProduct) {
      throw new Error("Product already exists!"));
    }
  } catch (err) {
    // ...
  }
}
```

Do you all agree? :)

sagirk:
@sbernaldev LGTM, with a nit! 🎉

nit: remove the extra `)`

```diff
- throw new Error("Product already exists!"));
+ throw new Error("Product already exists!");
```

Good work! Please go ahead and commit your changes. I can't wait to land it ASAP! 🥇

i0natan:
@sagirk @sbernaldev Great collaboration toward an improved code. 

Two thoughts from my part:
1. Important: This bullet had a major flaw which we might want to fix now. It's absolutely now a best practice to throw an error with description only as it doesn't provide the catcher with enough information to make thoughtful decisions (e.g. payment screen that must have as much information to avoid losing the customer). An error object must contain additional properties that classify the error - [even the Node core recently added 'code' property to all errors ](https://medium.com/the-node-js-collection/node-js-errors-changes-you-need-to-know-about-dc8c82417f65)

We already have an example for that at the bottom but the upper example (the one you improved) lacks this context. Could be great if you improve that, I wouldn't complicate the upper example too much (e.g. throw new myCustomError(404, "desc", errors.resourceDoesntExist //myCustomError derives from Error)

2. Minor: I personally prefer:
`If(product===null)//DAL documentation states that it returns null when resource doesnt exist`

Over

if(!product)//implicit check, JS checks many things like if it equals to zero

sbernaldev:
Hello @i0natan I try to do my best! Are you trying to find a solution like this? Should I keep _catch_ right?

```javascript
// 'throwing' an Error from a Promise
const addProduct = async (productToAdd) => {
  try {
    const existingProduct = await DAL.getProduct(productToAdd.id);
    if (existingProduct) {
      throw new AppError(409, "Product id already exists, the product has not been added", true);
    }
  } catch (err) {
    // ...
  }
}
```

Maybe the title should be another PR, in order to thread different tasks, what do you think?

sagirk:
@i0natan My thoughts (mostly to do with ensuring a friction-free contribution experience):

1. As far as regards the first suggestion, it increases the scope of what is to be changed quite a bit. I think we should create a new issue and address related changes on a new PR.

2. The second suggestion can be serviced in this PR. I can merge this PR in after the following change:

```diff
const addProduct = async (productToAdd) => {
  try {
    const existingProduct = await DAL.getProduct(productToAdd.id);
-   if (existingProduct) {
+   if (existingProduct !== null) {
      throw new Error("Product already exists!"));
    }
  } catch (err) {
    // ...
  }
}
```

@sbernaldev Let me know if you need any assistance making the second change. 

Also, if you are interested in taking up the first change, do you mind raising a follow-up PR to this? We can have further discussions on there.

sbernaldev:
I will resume this on monday-tuesday, I'm very pleasant but just almost stressed in a entrepreneur tournament!

i0natan:
@sbernaldev LGTM, anything you wanna change before I merge?

Shall we move then to make it even better by explaining that errors should be tagged?

@BrunoScheufler Wanna use this bullet, which is now renovated, to experiment TS or code standards linting? e.g. you may create an equivalent TS code example nearby each vanila-JS example

BrunoScheufler:
@i0natan will look into that, great idea!

i0natan:
@sagirk @BrunoScheufler Had to handle merge conflicts, your review was dismissed, kindly review again😞

sagirk:
Thanks for the contribution, @sbernaldev! 🥇

Issue #247, opened 2018-09-16
New practice suggestion: Prevent Unsafe Redirects


i0natan:
A great addition to our security best practices list:
https://blog.hailstone.io/how-to-prevent-unsafe-redirects-in-node-js/

js-kyle:
Merged :)

PR #262, opened 2018-10-07
MINOR Prevent unsafe redirects

js-kyle:
https://github.com/i0natan/nodebestpractices/issues/247

### DIFF
--- a/sections/security/saferedirects.md
+++ b/sections/security/saferedirects.md
@@ -1,0 +1,58 @@
+# Prevent unsafe redirects
+
+### One Paragraph Explainer
+
+When redirects are implemented in Node.js and/or Express, it's important to perform input validation on the server-side.
+If an attacker discovers that you are not validating external, user-supplied input, they may exploit this vulnerability by posting specially-crafted links on forums, social media, and other public places to get users to click it.
+
+Example: Unsafe express redirect using user input
+```javascript
+const express = require('express');
+const app = express();
+
+app.get('/login', function (req, res, next) {
+
+  if (req.session.isAuthenticated()) {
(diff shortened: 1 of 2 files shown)


sections/security/saferedirects.md
```diff
@@ -0,0 +1,58 @@
...
+When redirects are implemented in Node.js and/or Express, it's important to perform input validation on the server-side.
+If an attacker discovers that you are not validating external, user-supplied input, they may exploit this vulnerability by posting specially-crafted links on forums, social media, and other public places to get users to click it.
+
+Example: Unsafe express redirect using user input
+```javascript
+const express = require('express');
+const app = express();
+
+app.get('/login', function (req, res, next) {
+
```
BrunoScheufler: I think we should use an arrow function here. This might be different from the previous approach, but I think it's a part of our style guidelines and will become important in the future :+1:

sections/security/saferedirects.md
```diff
@@ -0,0 +1,58 @@
...
+  if (url.match(/^\/(?!\/)/)) { 
+    // Prepend our domain to make sure 
+    return 'https://example.com' + url; 
+  } 
+    // Otherwise check against a whitelist
+  return whitelist[url] ? url : '/'; 
+}
+
+app.get('/login', function (req, res, next) {
+
```
BrunoScheufler: This part would need arrow-functions as well :smile:

js-kyle:
@BrunoScheufler & @sagirk thanks for review - have updated 2x examples to use arrow functions, please merge if all is ok now! Cheers

i0natan:
@js-kyle Great addition

PR #263, opened 2018-10-08
Fix links

duartemendes:


BrunoScheufler:
@duartemendes thanks so much 💯

Issue #248, opened 2018-09-16
new practice suggestion: 


i0natan:
> Logging is one of those things that can sometimes be a black box for developers. Maybe you have a dedicated DevOps person who takes care of logging infrastructure for you, or maybe it’s your first time working on this side of things.

https://www.coreycleary.me/why-should-your-node-js-application-not-handle-log-routing/

i0natan:
@coreyc That's cool that you're here! Goes without saying that I enjoyed the article. How did you find this thread?

Looking forward to reading the PR. May I suggest that you take a quick look at our [writing guideliness](https://github.com/i0natan/nodebestpractices/blob/master/.operations/writing-guidelines.md)

coreyc:
@i0natan I noticed some traffic coming to my blog from this link (I have very basic analytics setup, so often it'll just show github.com as the referrer link but this time it showed the full url). Also, I'm well aware of this project so I'm sure I would have stumbled upon this thread sooner or later!

js-kyle:
This has been merged, thanks @coreyc !

PR #272, opened 2018-10-27
Add log routing best practice

coreyc:
First pass at this, please let me know your comments.

Also, the best practice suggested here slightly conflicts with what is presented in the ["smart logging" best practice](https://github.com/i0natan/nodebestpractices/blob/master/sections/errorhandling/usematurelogger.md).

In that example, the application does write to a file in the Winston example, and in the StrongLoop blog reference seems to suggest multiple log targets are fine (which they are!) but doesn't clarify if that's from the application or from the execution environment.

What do you think is the best way to consolidate the two?


js-kyle:
Hey @coreyc this looks great, I will give this a full review soon!

coreyc:
references issue #248

i0natan:
@coreyc @js-kyle Planning to review soon

coreyc:
@js-kyle thanks for reviewing! I fixed the typo and added the TLDR to the README.md

README.md
```diff
@@ -649,6 +649,16 @@ All statements above will return false if used with `===`
 
 🔗 [**Read More: Use an LTS release of Node.js**](/sections/production/LTSrelease.md)
 
+<br/><br/>
+
+## ![✔] 5.18. Log Routing
```
i0natan: Most of our bullets start with verb/action.

Like 'Don't route logs within the app', or 'Don't write to log files, prefer stdout' (the last sounds very intrigue as many people would be surprised)

README.md
```diff
@@ -649,6 +649,16 @@ All statements above will return false if used with `===`
...
+
+**TL;DR:** Logs should be routed not by the application, but by the exection environment the application runs in. Node applications typically write logs to `stdout` using the `console` module. Instead of the application specifying where the logs should end up, write to `stdout` using `console` and then let the execution environment (container, server, etc.) pipe the `stdout` stream to the appropriate destination (i.e. Splunk, Graylog, etc.).
```
i0natan: Maybe we should clarify *who should do that. The word "Application" might mislead as also the devops configuration is part of the application. Consider changing to "Logs destination should not be routed hard-coded by the developer, rather..."

If we are mentioning providers - elastic search/kibana are the most popular one

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+
+We often think of "separation of concerns" in terms of pieces of code between services and between services themselves, but this applies to the more “infrastructural” components as well. Your application code should not handle something that should be handled by infrastructure/the execution environment (most often these days, containers). What happens if you define the log locations in your application, but later you need to change that location? That results in a code change and deployment. Also, when working with container-based/cloud-based platforms, containers can spin up and shut down when scaling to performance demands, so we can't be sure where a logfile will end up. The execution environment (container) should decide where the log files get routed to instead. The application should just log what it needs to to `stdout` / `stderr`, and the execution environment should be configured to pick up the log stream from there and route it to where it needs to go.
```
i0natan: Consider emphasizing that this separation of concern is needed because the role who need to adjust this setting (log target) are sometimes not developers and we prevent them from doing this (it's not just about software aesthetics but real production problem)

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+
+Benefits:
```
i0natan: I would remove the last part - it's not anymore a single paragraph and the point is clear

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+- keeping the log routing responsibility out of your application code:
+    - keeps the code cleaner
+    - makes log routing locations easier to change without deployments
+- scaling applications/containers means it’s harder to have control over logfiles
+- scaling applications also means they’re more ephemeral, meaning logfiles might not be there depending on the state of the container
+- writing to, say a file or database, over stdout/stderr ties you down to those log targets, takes away your flexibility to pipe the output of stdout/stderr to whatever targets you want, and change this at the infrastructure or container level
+
+<br/><br/>
+
+### Code Example – Log routing tightly coupled to application
```
i0natan: Put some bold word like "Anti-pattern"

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+
+```javascript
+const { createLogger, transports, winston } = require('winston');
+const winston-mongodb = require('winston-mongodb');
```
i0natan: I would use a file because that's what most developers do. Should someone look at this example he might think "oh well of course I don't save logs to DB!". But that's not the point, even saving to log files is wrong which is quite a popular technique developers tend to embrace

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+
+From the [O'Reilly blog](https://www.oreilly.com/ideas/a-cloud-native-approach-to-logs),
+ > When you have a fixed number of instances on a fixed number of servers, storing logs on disk seems to make sense. However, when your application can dynamically go from 1 running instance to 100, and you have no idea where those instances are running, you need your cloud provider to deal with aggregating those logs on your behalf.
+
+<br/><br/>
+
+ ### Example: Architecture overview using Docker and Splunk as an example
+
+![alt text](/assets/images/logging-overview.png "Log routing overview")
+
```
i0natan: Maybe quote from 12 factor as well?

sections/production/logrouting.md
```diff
@@ -0,0 +1,76 @@
...
+winston.add(winston.transports.MongoDB, options);
+```
+Doing it this way, the application now handles both application/business logic AND log routing logic!
+
+<br/><br/>
+
+### Code Example – Better log handling + Docker example
+In the application:
+```javascript
+console.log('something we want to log here') // this writes to stdout
```
i0natan: Important: I wouldn't recommend direct logging to console.log rather going through a logger library like winston because we will benefit other important features like log levels

i0natan:
@js-kyle @js-kyle I'll be glad to spare another working hands. Some of my comments are just ideas and absolutely not mandatory changes

coreyc:
@i0natan @js-kyle thanks for both of your comments and reviews! I don't think it will take me long to make the changes, but if I have any points of clarification I'll make sure to post here.

I think the idea of still using a logger (like Winston) rather than console.log/err is good as it helps reach "parity" with the section currently in place on logging.  

Should only take me 10 min or so to incorporate changes where I see fit. And I've got some time tomorrow afternoon to make them!

README.md
jarzuaga: Small type `but by the exection` should be `but by the execution`

sections/production/logrouting.md
coreyc: Just to clarify, by "last part" you mean the "Benefits" bulleted list section?

sections/production/logrouting.md
coreyc: Yeah log files does seem to be pretty common. I have actually seen people logging to a database on several applications I've worked on though!  If you're ok with it I think I'm going to leave both (files + mongo) there.

coreyc:
Ok, I _think_ that should take care of the feedback. Thanks again for everyone's thorough comments. 

To be consistent, I did end up removing the File transport from the existing usematurelogger.md. Let me know what you think about that.

i0natan:
@js-kyle The ball is in your hands

@coreyc Consider adding a [pino example (maybe in the future)](https://github.com/pinojs/pino/issues/538)

PR #288, opened 2018-11-25
Added initial compression copy

TheHollidayInn:
Here is my initial draft for using compression. It's definitely rough, so comments are most welcome :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -951,9 +951,20 @@ All statements above will return false if used with `===`
 
 <p align="right"><a href="#table-of-contents">⬆ Return to top</a></p>
 
-# `Performance Practices`
+# `7. Performance Practices`
 
 ## Our contributors are working on this section. Would you like to join?
+
+## ![✔] 7.1 Use compression on your responses
+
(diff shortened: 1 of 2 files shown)


i0natan:
@TheHollidayInn I suggest to include this item as part of our generic bullet (each section has a generic bullet "Other advice") and not as an independent one because:

- We have a recommendation in our guide not to store static assets in Node. It's not common and usually slower to compress API responses
- If you do host static assets, then it's usually recommended to perform the compression outside of Node
- It's not an advice that relates specifically to Node rather to any web platform

Thoughts?

TheHollidayInn:
Yeah, I think that sounds good! This one is small as well.

TheHollidayInn:
Should I create the generic section?

i0natan:
> Should I create the generic section?

@TheHollidayInn Good idea, just keep the same formatting as 6.5 (security generic). Also, before that let's open an issue and put together few ideas (it doesn't make sense to open a generic bullet for 1 item)

PR #289, opened 2018-11-26
Chinese translation to lintrules

only-lei:


i0natan:
@MattJin Awesome!

I see many changes in non-Chinese files, is this intentional?

MattJin:
@i0natan  Yes. The Chinese branch was outdated, so we merged the changes from master branch.

PR #293, opened 2018-11-27
Updated outdated reference to nsp

ISNIT0:
See https://github.com/dominictarr/event-stream/issues/116#issuecomment-442181579

PR #295, opened 2018-11-30
Updated references to NSP

ISNIT0:
See https://github.com/i0natan/nodebestpractices/pull/293

I can't even pretend to read Chinese, does this change make sense in the Chinese version?


i0natan:
@ISNIT0 Great contribution

PR #300, opened 2018-12-14
Fix typo

reod:
agains -> against
missing dot at the end of sentence


BrunoScheufler:
@i0natan I've only assigned November flowers for now, every contribution in December will be done in the December update 👍

PR #312, opened 2019-01-01
fix npm link

justjavac:


BrunoScheufler:
@justjavac thanks for your contribution 💯

i0natan:
@BrunoScheufler How do we remember to assign a flower to @justjavac? should be some simple way to achieve this

cc @js-kyle, January assignee

Issue #315, opened 2019-01-05
Performance List of Generic Bullet points


TheHollidayInn:
Continuing from: https://github.com/i0natan/nodebestpractices/pull/288#issuecomment-450572627. As suggested, let's discuss a list of bullet points to be added to our generic list.

Looking at this: https://github.com/i0natan/nodebestpractices/issues/302, here are a few that seemed to be good:
- Use text compression
- Deal with database and external APIs in batches
- Benchmark slow db calls
- Analyze repeated API Calls/Database Queries to cache them
- Serve static assets from a CDN

Are these good? and what else should we add?

Also, the format will be similar to this: https://github.com/i0natan/nodebestpractices#-65-collection-of-generic-security-best-practices

i0natan:
@TheHollidayInn Perfect 🎖 

Just try adding some more info & links to each (when the implementation tools are highly popular and standard), for example:
"Benchmark slow db calls" - share link to Postgres & mySQL query plans and slow queries profiling

Let's start refining this using a PR?

i0natan:
@TheHollidayInn Didn't you have another performance PR? I suddenly can't find it...

TheHollidayInn:
Will do! I'll open a PR and add more.

Here is the other part I was working on: https://github.com/i0natan/nodebestpractices/issues/294#issuecomment-450262446. I posted the initial thoughts in the issue per your request. I can switch it to a PR if you'd like!

i0natan:
@TheHollidayInn Great, so I'm following on  #320 & #294

i0natan:
Closing as we're engaged on the PR #320

PR #320, opened 2019-01-13
[WIP] Added initial common performance list

TheHollidayInn:
This is a work in progress addressing our generic performance list: https://github.com/i0natan/nodebestpractices/issues/315

- [x] Use text compression
- [x] Deal with database and external APIs in batches
- [x] Benchmark slow db calls
- [x] Analyze repeated API Calls/Database Queries to cache them
- [x] Serve static assets from a CDN
- [ ] Use multiple level of caching (http, CDN, memory store, others?) **This may be covered above**
- [ ] Prepare the data for frequent queries in advance (have to find this technique name) **This may be covered above**
- [ ] Use a read DB replica to reduce load on the main DB tier (links to Mongo, Postgres replica doc)


js-kyle:
Looking like a great start

BrunoScheufler:
@i0natan I think it will depend on the content depth of the final list item contents 🤔 If it stays generic, we might as well just add a flat list as you suggest, otherwise the current structure might be better 👍

i0natan:
@TheHollidayInn "I'm happy to keep a flat format. That does seem much easier." Agree 👍 👍 . This section should be based on links - anytime a user wants to get more info he will just click 

Thoughts for new performance items:
- Use multiple level of caching (http, CDN, memory store, others?)
- Prepare the data for frequent queries in advance (have to find this technique name)
- Use a read DB replica to reduce load on the main DB tier (links to Mongo, Postgres replica doc)
- Others will follow soon...

TheHollidayInn:
Awesome! "Prepare the data for frequent queries in advance" is this not caching? 

For "Use multiple level of caching (http, CDN, memory store, others?)" we currently have a bullet for CDN, should I added a separate one for HTTP and Memory Store (Varnish/Redis) or combine them?

TheHollidayInn:
@i0natan Do you have any links for the items you suggested that I could read/reference. Also, two of your suggestions may be covered by above bullet points but let me know if not.

i0natan:
@TheHollidayInn As of preparing data - I didn't mean caching, rather than store the query data ready for serving instead of fetching it live. See the section '(1) Denormalization' here:
https://highlyscalable.wordpress.com/2012/03/01/nosql-data-modeling-techniques/

Other ideas: 
- Use CDN and etag to store assets that rarely change (e.g. countries list)
- Pool db connection (supported by all the major node.js ORMs)
- Compile DB queries in advance (include a link to mySQL, Postgres papers on this)

Could be great if you skim through papers about db/web/cloud performance practices

TheHollidayInn:
Ah okay. Yeah Denormalization is here to: https://github.com/donnemartin/system-design-primer/blob/master/README.md#denormalization.

Actually, if we are speaking about performance in general, I can add the above items you listed along with another bullet about using the System Design Primer. It has a nice section about Performance versus Scale:
```
A service is scalable if it results in increased performance in a manner proportional to resources added. Generally, increasing performance means serving more units of work, but it can also be to handle larger units of work, such as when datasets grow.1

Another way to look at performance vs scalability:

    If you have a performance problem, your system is slow for a single user.
    If you have a scalability problem, your system is fast for a single user but slow under heavy load.
```

And lots of resources on how to check your system, find bottleknecks and expand. They are very similar if not inclusive of the points above. I use the reference myself for many projects even non-Node ones.

i0natan:
I wasn't aware of this repo, looks amazing, I must spend time with it :]

TheHollidayInn:
Any updates for this?

i0natan:
@TheHollidayInn In few days will bounce with updates

/remind me in 3 days

Calinou:
It might be worth it to mention Brotli compression as well. It usually performs better than gzip (especially on static assets which can be compressed at build-time) while being [well-supported by modern browsers](https://caniuse.com/#feat=brotli).

Issue #321, opened 2019-01-14
Use images w/ transparent background – dark theme FTW


scorphus:
First of all, thanks for these best practices!

Just like many others, I use a [dark theme](https://github.com/StylishThemes/GitHub-Dark) for GitHub with [Stylus](https://github.com/openstyles/stylus/):

![image](https://user-images.githubusercontent.com/2112697/51123966-75252e80-181d-11e9-9baa-12068c3c75b5.png)

![image](https://user-images.githubusercontent.com/2112697/51124342-655a1a00-181e-11e9-847b-cbf38f9e8c94.png)

This amazing best practices would look a bit better on dark themes if those images had transparent background.

I'd love to change that and submit a PR. And it would be much easier if I had the original image files, if that's even possible.

js-kyle:
Hey @scorphus - cool idea, a PR for this would be welcomed! I'll try to find out if those files are available before you start

i0natan:
@scorphus Sounds interesting, can the user opt-in/out? where can I see an example? Kindly elaborate how it works

scorphus:
> @scorphus Sounds interesting, can the user opt-in/out? where can I see an example? Kindly elaborate how it works

I'm not sure I get it, but I'll try to elaborate. Pardon me if I got it wrong. An image with transparent background blends itself with the background it's on top of.

In theory, it could be seen as a kind of opt-in/out. That is, depending on what background colour the user defines, the image will have that background colour. But in reality, no. The image will be the same for every single user.

[In this example](https://github.com/scorphus/nodebestpractices/tree/banner-transp-bg), the banner has transparent background.

By the way, we can start with the banner: #327

Thanks!

i0natan:
Got it, so we basically need to make the banner and other images transparent?

i0natan:
@idori Do we have those?

idori:
@i0natan yes, I'll re-make the header with a transparent background and check which other images require work to be dark mode friendly

i0natan:
Thanks @idori

PR #327, opened 2019-01-19
Add a transparent background to the banner image

scorphus:
Add a new PNG image with transparent background and use it as the default banner on README.md.

Besides looking better on dark themes, the new new banner is around 90% smaller than the one it replaces.

See #321 for reference.


i0natan:
@js-kyle Not sure what exactly to review, there is a new header image, that's it? how one can see the site in dark mode? Please instruct a layman

idori:
I've recreated the original header with a transparent background. Need some finishing touches with few other small images and I'll submit a PR

i0natan:
@idori Shall we wait for other edits or approve this?

PR #336, opened 2019-02-14
typo fix at 4.1

ArlindXh:
Small typo fix in the TL;DR of 4.1.


ArlindXh:
Agreed, I changed the "provide" back to "provides".  I think this makes more sense.

BrunoScheufler:
@ArlindXh thanks so much for your contribution 🎉

PR #342, opened 2019-02-20
Include new Function within avoideval.md

LewisArdern:
Dynamic execution of JavaScript also happens in new Function, which is even referenced here: https://github.com/i0natan/nodebestpractices#-615-avoid-javascript-eval-statements, so this should be included.


js-kyle:
Thank you @LewisArdern !

PR #347, opened 2019-02-25
Updates to commonsecuritybestpractices.md

LewisArdern:
Did not like that OWASP A7 only included 'use CSP, so updated that to reference more content from OWASP with some slight changes to reflect node, e.g use a template engine.

Made some slight changes to the content for grammar improvements


js-kyle:
@LewisArdern really like + appreciate your attention to detail here and on other PR's, thanks!

i0natan:
@LewisArdern Great addition

PR #349, opened 2019-03-05
fix markdown error

js-kyle:


js-kyle:
The markdown is breaking without this new line - could someone merge this please? It needs an approving review before I can do that

PR #362, opened 2019-03-17
Added stale robot that will auto-close silent issues

i0natan:


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
...
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - security
+# Label to use when marking an issue as stale
+staleLabel: stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
```
sagirk: nit: Maybe sprinkle a bit of 'human' flavor over it? 😅

```
>
  Hello there! 👋
  
  This issue has gone silent. Eerily silent. ⏳
  
  We currently close issues after 100 days of inactivity. It has been 90 days
  since the last update here.
  
  If needed, you can keep it open by replying here.
  
  Thanks for being a part of the Node.js Best Practices community! 💚
```

BrunoScheufler: Sounds like a good idea, let's make it sound friendlier 👍

Issue #172, opened 2018-04-11
April maintenance list


i0natan:
Assignee: @i0natan

**Updates**
* Update top badges with Node version and last update date
* Ensure all translations are aligned with the English version

**Flowers**
@keyfoxth
@utkarshbhatt12 

**Stars**

sagirk:
It seems we're behind on the April maintenance list.

Also, have we not opened a May maintenance list yet?

@i0natan @idori @refack @BrunoScheufler I'd like to help out, if that's okay.

sagirk:
Update: Filed an issue for [May 2018 maintenance](https://github.com/i0natan/nodebestpractices/issues/183).

BrunoScheufler:
@sagirk yeah you're right, if you're up to adding the flowers for this list and create a PR, that would be helpful!

i0natan:
@sagirk Your help is much appreciated

sagirk:
Opened a [PR](https://github.com/i0natan/nodebestpractices/pull/185).

@i0natan @BrunoScheufler 
Taken care of everything for April except this bit:
* Ensure all translations are aligned with the English version

What are the steps to ensure above?

BrunoScheufler:
This is rather complicated due to the need of staying in contact with the translation groups, don't worry, we'll happily take care of that. Great job however!

i0natan:
@sagirk If you're up to additional improvements, we've bunch of improvement ideas in our bag waiting for someone to push them forward

BrunoScheufler:
@i0natan do we? 😅

i0natan:
:)

Yeah:
1. Bibliography project - extracting all the sources we used into one list on the home page including books, bloggers, etc.
2. For each bullet, show list of top vendors with a comparison of the key metrics (e.g. monitoring solutions)
3. Others

I didn't ask from you to look into one of this because you seem busy but should you have time I'd be glad if you can handle some additional project

BrunoScheufler:
Oh no worries 😄 I'll message you if I have the demand for these tasks!

sagirk:
@i0natan I can start with the Bibliography project. Where can I learn more? Any existing issue/repo/example?

Just to be clear, this is different from the planned [credit section](https://github.com/i0natan/nodebestpractices/issues/90), right?

i0natan:
[...]

PR #365, opened 2019-03-25
Rename 3.4 Use Semicolon to proper statement separation

### LINKED ISSUE #26: 3.4 Don't Forget the Semicolon
Hi

I saw your section "3.4 Don't Forget the Semicolon"

Don't forget that the semicolon "cannot save you".

Indeed, look at this code:

````javascript
function aaa() {
    return
    {
        test: 1
    };
}
aaa()
````

If you run it, you will have ... undefined !!
Because the "{" of the json response is not on the same line of the return statement.

So, prefer using Eslint with adapted rules to avoid this kind of issue.

A cool video too to show more: https://www.youtube.com/watch?v=Qlr-FGbhKaI

Regards

### PR DESCRIPTION
**Pull request by MartinMuzatko**
Fixes #26.

I hope I integrated all feedback from this issue, while leaving the core message intact: People should be aware how statements are separated or linked together, and that there are cases where ASI and/or semicolon do not help, if statement termination is not understood. This also leaves room for the reader to prefer semicolons or not.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ function someFunction()
 
 No matter if you use semicolons or not to separate your statements, knowing the common pitfalls of improper linebreaks or automatic semicolon insertion, will help you to eliminate regular syntax errors.
 
-**TL;DR:** Use ESLint to gain awareness about separation concerns. Prettier or Standardjs can automatically resolve these issues.
+**TL;DR:** Use ESLint to gain awareness about separation concerns. [Prettier](https://prettier.io/) or [Standardjs](https://standardjs.com/) can automatically resolve these issues.
 
 **Otherwise:** As seen in the previous section, JavaScript's interpreter automatically adds a semicolon at the end of a statement if there isn't one, or considers a statement as not ended where it should, which might lead to some undesired results. You can use assignments and avoid using immediate invoked function expressions to prevent most of unexpected errors.
 
--- a/README.md
+++ b/README.md
@@ -273,9 +273,9 @@ function someFunction()
 
 ## ![✔] 3.4 Separate your statements properly
 
-No matter if you use semicolons or not to separate your statements, knowing the common pitfalls of inproper linebreak or automatic semicolon insertion, will help you to eliminate regular syntax errors.
+No matter if you use semicolons or not to separate your statements, knowing the common pitfalls of improper linebreaks or automatic semicolon insertion, will help you to eliminate regular syntax errors.
 
-**TL;DR:** Use ESLint to make you aware about separation concerns. Prettier or Standardjs can automatically resolve these issues.
+**TL;DR:** Use ESLint to gain awareness about separation concerns. Prettier or Standardjs can automatically resolve these issues.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: README.md [hunk 1]
```diff
@@ -271,28 +271,48 @@ function someFunction()
...
 <br/><br/>
 
-## ![✔] 3.4 Don't Forget the Semicolon
+## ![✔] 3.4 Separate your statements properly
 
-**TL;DR:** While not unanimously agreed upon, it is still recommended to put a semicolon at the end of each statement. This will make your code more readable and explicit to other developers who read it
+No matter if you use semicolons or not to separate your statements, knowing the common pitfalls of inproper linebreak or automatic semicolon insertion, will help you to eliminate regular syntax errors.
 
-**Otherwise:** As seen in the previous section, JavaScript's interpreter automatically adds a semicolon at the end of a statement if there isn't one, or considers a statement as not ended where it should, which might lead to some undesired results
+**TL;DR:** Use ESLint to make you aware about separation concerns. Prettier or Standardjs can automatically resolve these issues.
```
BrunoScheufler: Maybe you could rephrase this into "...to make yourself aware..." or "...to gain awareness..." but that's just a minor thing 👍

README.md
```diff
@@ -271,28 +271,48 @@ function someFunction()
...
 
 ### Code example
 
 ```javascript
 // Do
-const count = 2;
-(function doSomething() {
-  // do something amazing
-}());
+function doThing() {
```
i0natan: @MartinMuzatko @BrunoScheufler 

Could we separate this into 2-3 very short and concise examples?

MartinMuzatko: I have the following suggestions, but would like to have your feedback:

**Do:** Define your used values in variables or function parameters
``` js
const items = [1, 2, 3]
items.map(item => item * 3)
```

``` js
function timesThree(numbers) {
    return numbers.map(number => number * 3)
}
```

**Avoid:** Unlogical order of statements - only define statements in an order as they are needed. If your code works in a side-effect free manner, then you would only define variables as late as they are needed.
``` js
const set = ['a', 'b', 'c']
[1, 2, 3].map(console.log)
// Uncaught TypeError: Cannot read property 'map' of undefined
```

Maybe you have suggestions for other examples?
As isaac states: https://blog.izs.me/2010/12/an-open-letter-to-javascript-leaders-regarding

>In general, \n ends a statement unless:
> 1. The statement has an unclosed paren, array literal, or object literal or ends in some other way that is not a valid way to end a statement. (For instance, ending with . or ,.)
> 2. The line is -- or ++ (in which case it will decrement/increment the next token.)
> 3. It is a for(), while(), do, if(), or else, and there is no {
> 4. The next line starts with [, (, +, *, /, -, ,, ., or some other binary operator that can only be found between two tokens in a single expression.

MartinMuzatko: Also I would maybe add:
> Clever short-hands are discouraged, in favor of clear and readable expressions, whenever possible.
> source: https://standardjs.com/rules.html#semicolons

i0natan:
@MartinMuzatko Thanks a bunch for the great contribution! This guide is now better thanks to the effort of the community. I'm assigning flyer here #172 and will Tweet this PR soon

PR #366, opened 2019-03-26
6.12 fix login rate limit - brute-force prevention

animir:
1. README add common technique description and remove non-maintained package link
1. Improve title and remove non-maintained package example

### DIFF
--- a/README.md
+++ b/README.md
@@ -820,7 +820,7 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 6.12. Limit the allowed login requests of each user
+## ![✔] 6.12. Prevent brute-force attacks against authorization
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A2-Broken_Authentication" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A9:Broken%20Authentication%20-green.svg" alt=""/></a>
 
--- a/README.md
+++ b/README.md
@@ -824,7 +824,10 @@ All statements above will return false if used with `===`
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A2-Broken_Authentication" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A9:Broken%20Authentication%20-green.svg" alt=""/></a>
 
-**TL;DR:** A brute force protection middleware such as [express-brute](https://www.npmjs.com/package/express-brute) should be used inside an express application to prevent brute force/dictionary attacks on sensitive routes such as /admin or /login based on request properties such as the username, or other identifiers such as body parameters
+**TL;DR:** A simple and powerful technique is to limit authorization attempts using two metrics:
+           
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -824,7 +824,10 @@ All statements above will return false if used with `===`
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A2-Broken_Authentication" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A9:Broken%20Authentication%20-green.svg" alt=""/></a>
 
-**TL;DR:** A brute force protection middleware such as [express-brute](https://www.npmjs.com/package/express-brute) should be used inside an express application to prevent brute force/dictionary attacks on sensitive routes such as /admin or /login based on request properties such as the username, or other identifiers such as body parameters
+**TL;DR:** A simple and powerful technique is to limit authorization attempts using two metrics:
+           
+1. The first is number of consecutive failed attempts by the same user unique ID/name and IP address.
+1. The second is number of failed attempts from an IP address over some long period of time. For example, block an IP address if it makes 100 failed attempts in one day.
```
js-kyle: Minor - this should be 2.

sections/security/login-rate-limit.md
```diff
@@ -1,51 +1,9 @@
-# Implement express rate limiting for login routes
+# Prevent brute-force attacks against authorization
```
js-kyle: It would make sense to also update the bullet title 6.12 on README.md to match this, so they are consistent

README.md
animir: This is markdown syntax - if you preview this file, it looks fine.

sections/security/login-rate-limit.md
animir: Done

README.md
js-kyle: Yes, but it is better for the source to match the output imo

animir: ah, right. fixed.

README.md
```diff
@@ -820,11 +820,14 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 6.12. Limit the allowed login requests of each user
+## ![✔] 6.12. Prevent brute-force attacks against authorization
```
BrunoScheufler: Maybe we could add "endpoints" to this headline, so "Prevent brute-force attacks again authorization endpoints", what do you think?

animir: Let's keep it as short as possible. This phrase has the same meaning with or without "endpoints" word in the end.

BrunoScheufler: I don't think it's complete in that case though. For the sake of readability I would vouch for adding endpoints, one additional word doesn't change the whole thing. But I'd like to get more opinions on this, what do you think @sagirk @i0natan @js-kyle ?

i0natan:
@BrunoScheufler @js-kyle Assigned to whom?

README.md
animir: @BrunoScheufler Not all frameworks/libraries use "endpoint" term. For example, developers name it **authorization resolver** in GraphQL. It may be just **authorization service** in service oriented architecture. If we do not add it, title would be wider.

Issue #294, opened 2018-11-29
Performance: Don't block the event loop


js-kyle:
TLDR: You should make sure you never block the Event Loop. Node is fast when the work associated with each client at any given time is "small".  In other words, each of your JavaScript callbacks should complete quickly, to avoid performance impacts. 

Otherwise: The throughput (requests/second) of your server will suffer.

Suggested resources: 

- https://nodejs.org/en/docs/guides/dont-block-the-event-loop/

TheHollidayInn:
Can I take this?

i0natan:
Ideas:

- Monitor the event loop block and bake into your monitoring
- Mitigation techniques when having CPU intensive tasks: run the heavy tasks offline, other 3-4 techniques here
- Use setImmediate to break long CPU-bounded tasks into multiple smaller tasks that wait in the queue for execution
- Code example for detection:
```
const blocked = require('blocked-at');

const greedyFunction = () =>{
    const end = Date.now() + 2000;
            while (Date.now() < end) {
            eval("1+2+3");
        } 
    console.log(`Ending a greedy function`);
}

const handleBlockedLoop = (time, stack) => {
    console.log(`Blocked at ${stack} for ${time} ms`);
    //TODO: export to Promotheus here
}

const {stop} = blocked(handleBlockedLoop, {});
setImmediate(greedyFunction);
console.log(`Program finished`);
```

Other ideas?

i0natan:
@TheHollidayInn 

Sure, with pleasure. Let's just first exchange some ideas here?

TheHollidayInn:
Yeah, definitely!

Few questions: 
- Do you have a recommended way to monitor? I'm guessing monitoring CPU usage/cycles would be helpful in tracking these down
- Should we recommend queue/worker system for offloading cpu tasks?
- "..JavaScript callbacks should complete quickly, to avoid performance impacts." can we add a snippet including ensuring your promises and awaited tasks end quickly - this is discussed in the link above.
- What do you think of including this express example: https://stackoverflow.com/questions/51495928/how-does-node-process-concurrent-requests. I think this is a common pattern where you may linger tasks after responding to the users requests

i0natan:
@TheHollidayInn My thoughts:
[...]

PR #373, opened 2019-04-09
Added initial event loop copy

TheHollidayInn:
Here is the initial *rough* draft for #294


i0natan:
@TheHollidayInn I will review this in the next few days. Looks like a great start!

TheHollidayInn:
@i0natan and updates on this?

i0natan:
@TheHollidayInn Sorry for the delay, I'm just struggling to jungle so many balls :)

If works for you, I can prioritize this bullet and run now a simulntanous discussion to finalize it next week.

In overall, I believe we should separate this into 3 bullets, sorted by priority:
1. Avoiding CPU heavy tasks - "Other CPU intensive tasks, such as DNS Lookup or Crypto will be offload to lubuv, take up other threads and starve new CPU intensive tasks"

2. Monitor the event loop

3. Avoid sync operations - "TLDR; Avoid using syncronous and other Event Loop blocking functions"

TheHollidayInn:
No problem! I'm happy to split/work on the three

i0natan:
@TheHollidayInn Cool, so here's my proposed TOC:

[TLDR Basics] The Node.js userland is single-threaded, keeping the CPU busy might put ~2000 concurrent users in idle. 
[TLDR spice it up a little] Yet, it's important to understand that most reasonable CPU operations, including simple cryptography and basic loops, barely affects the end user latency. One has iterate over arrays of million items to generate a noticeable block. Mitigating this demands no more than a simple monitor and developer awareness
[Otherwise] If only one request is doing some crazy computations (e.g. 100 crypto methods per request), all the other 1000 will wait patiently for that one to complete

[One paragraph] 
- Two lines about the event loop model - our app code is executed using a single thread (as opposed to the IO calls that go outside), this is great for minimizing context switching but assume that all behave nicely as they share the same...

[Examples & code]
- Diagram of the event loop that illustrates the block
- 7 lines code example that shows loop of crypto and a setTimeout that is called-back after 5 seconds instead of 2 (settime is set to 2000 ms but it will never called before the crypto part ends)
- Some nice quote
- Benchmark! I have this ready, Express route with 1...1000 loop, if I approach it using 1000 concurrent users, the loop add 3ms latency to the end-user (negelectible)

What do you think?

TheHollidayInn:
Hmm. Rethinking this and I might be missing how to separate all three. The issue with using CPU intensive tasks is that they are sync and block the event loop. I think all bullet points might be the same.

Taking a look at your copy now. Also, have you had a chance to read through my copy in this PR?

i0natan:
@TheHollidayInn 

Yes, I read your copy which was good but felt like we mix concerns, hence we need to separate, 
I don't see how the 3 are the same. 

1. Avoid CPU intensive tasks - this issues is related to code that is CPU intensive like crypto, see the proposed TOC

2. Is about monitoring it...

3. sync operations -  When you use fs.readfilesync then there's no CPU work here at all, it just blocks your loop until the **IO** work is over. That's a very specific way of blocking the loop and hurting performance. Very few people do that

Thoughts?

TheHollidayInn:
Ah okay, after reading everything, I am seeing what you mean. 

I think where I am getting confused is that both IO and CPU operations cause the same problem which is blocking the loop. And both are run on worker threads: https://nodejs.org/en/docs/guides/dont-block-the-event-loop/#what-code-runs-on-the-worker-pool

So in general the idea is to watch for any loop blockers - dns, file io, crypto, encryption, etc. 

But there are quite a few to watch for. I'd agree to separate them up for sure. I'll go ahead and write those up :D

i0natan:
@TheHollidayInn 

Sounds good, I suggest starting with 1 bullet, see my proposed TOC for the 1st one. Does this TOC resonate with you?

Minor technical remark:

> "And both are run on worker threads"

Most CPU intensive tasks won't run on worker thread, any user code except some Crypto/DNS libraries will get executed within the main thread. This article refers to the Node core code, not to the user-land code

TheHollidayInn:
@i0natan Yep! Works for me :D

I'm not sure I follow the difference between your distinction of node core and user-land. Crypto is considered CPI intensive tasks. DNS is IO. Both go to the worker pool. The section I linked refers to apis you can use in node modules.

I'm sure you can make something cpu intensive that is not listed, so I agree. However, your first TOC is "Avoid CPU intensive tasks - this issues is related to code that is CPU intensive like crypto, see the proposed TOC". Are you referring to the cypto module: https://nodejs.org/api/crypto.html - because the cpu intensive tasks from that module are run on the Worker Pool: `crypto.pbkdf2(), crypto.scrypt(), crypto.randomBytes(), crypto.randomFill(), crypto.generateKeyPair()`

Can you give me an example of a CPU tasks you are thinking?

i0natan:
> So I guess my new thought is there are operations to avoid that are cpu expensive - not on worker threads:
> 
> * algorithms with large number of operations (like your while loop)
> * Large json parsing
> * Certain regex queries
> * long promises/awaits

Yes! Those examples make this bullet shine. Crypto is just another example.

Now let's discuss Crypto which is just another bullet:
- Crypto work (except some built-in crypto utils that are executed using the libuv threads pool)

Only few crypto libraries, a subset of those that are part of Node, are executed using the thread pool, this is why crypto is still a good example

Synchronous APIs are a different bullet, keep it focused on a single challenge - CPU. Because soon you'll have to remind solutions and mitigating CPU intensive tasks take a different form than mitigating sync-IO tasks (which are very eay to remedy - just avoid them :))

Makes sense?

TheHollidayInn:
Alright, got a new draft up. Can you let me know what you think?

README.md
```diff
@@ -955,7 +955,17 @@ All statements above will return false if used with `===`
 
 ## Our contributors are working on this section. [Would you like to join?](https://github.com/i0natan/nodebestpractices/issues/256)
 
-<br/><br/><br/>
+<br/><br/>
+
+## ![✔] 7.1. Don't block the event loop
+
+**TL;DR:** TLDR; Avoid CPU intensive tasks as they will block the Event Loop.
```
i0natan: Duplicate TLDR?

Maybe add "block the mostly single-threaded event loop"?

sections/performance/block-loop.md
```diff
@@ -0,0 +1,57 @@
+Node handles the Event Loop on a single thread rotating through multiple queues. Operations with high complexity, large json parsing, unsafe regex querieies, and long awaited promises are some of the operations that can cause the event loop to stall. Preventing this is as easy as setting up monitoring for the the Event Loop and resolving the blocker.
```
i0natan: Did you forget the page title? see other 'more info' files, they all contain a title

i0natan: "handles the Event Loop on a single thread " -> "mostly on a single thread" (just to be precise)

"Operations with high CPU cost" (it's not about complexity, IO can be too complex)

"querieies" - typo?

"long awaited promises" - what do you mean here? not sure this is in-general true, usually promises are used to wait for IO

"Preventing this is as easy as setting up monitoring for the the Event Loop and resolving the blocker." - monitoring is detection, not prevention. Let's keep the monitoring to a dedicated bullet. My thoughts "Avoid this off-loading CPU intensive tasks to dedicated workload that doesn't serer users (e.g. job server), or by breaking long tasks into small steps that stand on the event queue lines, by using crypto libraries that utilize the event loop thread pool. To name a few techniques that can keep the event loop alive"

sections/performance/block-loop.md
```diff
@@ -0,0 +1,57 @@
...
+
+server.get('/', function (req, res, next) {
+  sleep(30)
+  res.send({})
+  next()
+})
+
+server.listen(3000)
+
+process.on('SIGINT', function () {
```
i0natan: Can you shorten this example by 50%? the SIGINT for example is not relevant to this topic

sections/performance/block-loop.md
```diff
@@ -0,0 +1,57 @@
...
+┌───────────┬─────────┬─────────┬─────────┬────────┬─────────┬───────┬─────────┐
+│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%  │ Avg     │ Stdev │ Min     │
+├───────────┼─────────┼─────────┼─────────┼────────┼─────────┼───────┼─────────┤
+│ Req/Sec   │ 31      │ 31      │ 33      │ 34     │ 32.71   │ 1.01  │ 31      │
+├───────────┼─────────┼─────────┼─────────┼────────┼─────────┼───────┼─────────┤
+│ Bytes/Sec │ 4.65 kB │ 4.65 kB │ 4.95 kB │ 5.1 kB │ 4.91 kB │ 151 B │ 4.65 kB │
+```
+
+## Image of the Event Loop
+![1*aU5dr98pxTsZ4AMfnA6lNA.png](https://cdn-images-1.medium.com/max/1600/1*aU5dr98pxTsZ4AMfnA6lNA.png)
```
i0natan: I'm not sure that this image is descriptive enough to explain why the loop is sensitive to CPU-bounded tasks. Can we find a better one?

sections/performance/block-loop.md
TheHollidayInn: A lot of my info comes from this article: https://nodejs.org/en/docs/guides/dont-block-the-event-loop/

From this section: https://nodejs.org/en/docs/guides/dont-block-the-event-loop/#don-t-block-the-event-loop

For promises: "You should make sure you never block the Event Loop. In other words, each of your JavaScript callbacks should complete quickly. This of course also applies to your await's, your Promise.then's, and so on."

For computational complexity
"A good way to ensure this is to reason about the "computational complexity" of your callbacks."

Also, I thought IO is going to be a separate bullet point. IO blocks the event loop for a different reason: 
```
I/O-intensive
    DNS: dns.lookup(), dns.lookupService().
    File System: All file system APIs except fs.FSWatcher() and those that are explicitly synchronous use libuv's threadpool.
```
If not, can you give me a counter example?

i0natan: > For promises: "You should make sure you never block the Event Loop. In other words, each of your JavaScript callbacks should complete quickly. This of course also applies to your await's, your Promise.then's, and so on."

Note there is a signficant difference between what you wrote and the article - you alluded to 'long awaited promises', they refer to the .then part. Practically, most promises wrapping some IO work, if you wait for long time then it's not related to CPU work. Their advice of not slowing down the .then handler makes sense, but it's too specific, we can generalize this advice to any other code...

> "A good way to ensure this is to reason about the "computational complexity" of your callbacks."

Well, it's probably right technically, is it a simple and well understood term? I guess the answer is subjective. I suggest that our guide should curate and simplify those technical guides

> For non-sync IO, should I state that you should also watch for large IO tasks?
 
I think it doesn't belong here
 
 

Note that I also included in my comment multiple mitigation techniques

TheHollidayInn: I'll definitely include many of your suggestions. I'm just trying to hone in on "high CPU usage" because it is super generic. I want to tell people what to look for.

I believe that these are three examples: `Operations with high complexity, large json parsing, unsafe regex queries`. So "high complexity" is just one example.

Also, promises are also used a lot for db access. If you have long queries that you are are `awaiting` you could block the loop like our example, I think.

Anyway, I'm just trying to give devs examples of "cpu work" to look for :D. I'll get your edits in this weekend.

TheHollidayInn: OKay, after reading this I'd retract my comment about `awaiting`. I was missing the idea that promises are added to the job queue: https://github.com/getify/You-Dont-Know-JS/blob/master/async%20%26%20performance/ch1.md. Soooo sorry about that :P

i0natan: Yes, if you await on IO operation (E.g. DB call) - this will not the event loop

TheHollidayInn:
@i0natan Can you review my updates :D

README.md
```diff
@@ -955,7 +955,17 @@ All statements above will return false if used with `===`
 
 ## Our contributors are working on this section. [Would you like to join?](https://github.com/i0natan/nodebestpractices/issues/256)
 
-<br/><br/><br/>
+<br/><br/>
+
+## ![✔] 7.1. Don't block the event loop
+
+**TL;DR:** Avoid CPU intensive tasks as they will block the mostly single-threaded Event Loop.
```
i0natan: You stated what NOT to do, let's include your recommendation? for example "Avoid CPU intensive tasks as they will block the mostly single-threaded Event Loop **and offload those to a dedicated thread, process or even a different technology based on the context**"

README.md
```diff
@@ -955,7 +955,17 @@ All statements above will return false if used with `===`
...
+
+**Otherwise:** As the Event Loop is blocked, Node.js will be unable to handle other request thus causing delays for concurrent users.
```
i0natan: Consider emphasizing the severity of this in a bolder language - "3000 users are waiting for a response, the content is ready to be served, but one single request blocks the server from dispatching the results back"

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
+# Don't block the event loop
+
+<br/><br/>
+
+Node handles the Event Loop mostly on a single thread rotating through multiple queues. Operations with high complexity, large json parsing, unsafe regex querieies, and large IO operations are some of the operations that can cause the Event Loop to stall. Avoid this off-loading CPU intensive tasks to a dedicated service (e.g. job server), or breaking long tasks into small steps then using the Worker Pool are some examples of how to avoid blocking the Event Loop.
```
i0natan: One more example "large JSON parsing, **applying logic over huge arrays**, unsafe regex..."

i0natan: Let's now take it one step up and make our advice even more accurate, I'll be glad to explain why is this true:

"When handling CPU-intensive tasks, different techniques should be applied based on of two different scenarios: if these CPU-intensive tasks are needed rarely (e.g. trigerred by a few requests) - then they can be offloaded to a dedicate process, thread or get handled in chunks (See example below). However, the more challenging case is where most requests need to trigger CPU-intensive code -  in that case, this task must be offloaded to a different workload, one that doesn't dispatch content to the end-users, like a job-server Serverless, and similar

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Example
```
i0natan: Formatting issue - in all other pages, the title contains the main message, e.g. "Example: blocking the event loop"

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
...
+Let's take a look at an example from [Node Clinic](https://clinicjs.org/documentation/doctor/05-fixing-event-loop-problem).
+```
+const restify = require('restify')
```
i0natan: Code examples should be really short and concise, the first two lines and the last one are not needed (they make it look long)

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
...
+│  ┌─────────────┴─────────────┐      └───────────────┘
+│  │           check           │
+│  └─────────────┬─────────────┘
+│  ┌─────────────┴─────────────┐
+└──┤      close callbacks      │
+   └───────────────────────────┘
+
+>Here's a good rule of thumb for keeping your Node server speedy: Node is fast when the work associated with each client at any given time is "small".
+>[Don't Block the Event Loop (or the Worker Pool) \| Node.js](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/)
+
```
i0natan: Try viewing the file - the formatting is broken, also see other pages for a reference

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
...
+while loop.
+
+### Run the benchmark 
+`clinic doctor --on-port 'autocannon localhost:$PORT' -- node slow-event-loop`
+
+### The results
+
+```
+─────────┬────────┬────────┬────────┬────────┬───────────┬──────────┬───────────┐
+│ Stat    │ 2.5%   │ 50%    │ 97.5%  │ 99%    │ Avg       │ Stdev    │ Max       │
```
i0natan: This is a lot of information - what is the bottom line? what takeaway do we want the user to extract from this?

sections/performance/block-loop.md
```diff
@@ -0,0 +1,73 @@
...
+┌───────────┬─────────┬─────────┬─────────┬────────┬─────────┬───────┬─────────┐
+│ Stat      │ 1%      │ 2.5%    │ 50%     │ 97.5%  │ Avg     │ Stdev │ Min     │
+├───────────┼─────────┼─────────┼─────────┼────────┼─────────┼───────┼─────────┤
+│ Req/Sec   │ 31      │ 31      │ 33      │ 34     │ 32.71   │ 1.01  │ 31      │
+├───────────┼─────────┼─────────┼─────────┼────────┼─────────┼───────┼─────────┤
+│ Bytes/Sec │ 4.65 kB │ 4.65 kB │ 4.95 kB │ 5.1 kB │ 4.91 kB │ 151 B │ 4.65 kB │
+```
+
+## Image of the Event Loop
+   ┌───────────────────────────┐
```
i0natan: This image is better, it gives a sense of the queues that might stay idle. Thinking out loud whether it's the best we can get? does it illustrate how some request blocks others? I tried to sketch something, see below, what do you think?

image

sections/performance/block-loop.md
TheHollidayInn: Sounds good to me and makes sense. Should I add this as a separate paragraph?

sections/performance/block-loop.md
TheHollidayInn: I think I've fixed this. When I view online or in VScode it looks normal.

sections/performance/block-loop.md
TheHollidayInn: This should be fixed now. Viewing in VScode or online works.

sections/performance/block-loop.md
TheHollidayInn: Basically this is out the output from node clinic showing the high latency, but I can see it not being helpful without the context show node clinic show the loop block issue.

TheHollidayInn:
I missed your comment but I'm happy to do got chat or hangouts :)

js-kyle:
Awesome work going on here! Nice

sections/performance/block-loop.md
i0natan: Yes, or maybe to the main one?

i0natan:
@TheHollidayInn Let me know when do you want me to review the whole, I noted that you answered some of the comments and other don't

sections/performance/block-loop.md
TheHollidayInn: @i0natan Do you have a suggestion on this section?

i0natan:
@TheHollidayInn Any chance that you're available for a 1-hour finalization session afterwhich, we will publish this? when might be a good time next week? 

Best if you can merge all my remarks into the PR by then so we just focus on final touches

i0natan:
@BrunoScheufler Can you review the lexical part + formatting?

TheHollidayInn:
Sweeeet!

@i0natan Do you have any time Tuesday or Thursday after 6pm UTC?

i0natan:
@TheHollidayInn What timezone are you in? my calendar seems busy at those timeslots, do you happen to be available on Friday 6-8am UTC or Sunday 12-16 UTC or Monday 12-16?

goldbergyoni:
@TheHollidayInn Do you have some availability this week and can proposes 2-3 slots?

TheHollidayInn:
Sorry! I actually had to do a last minute travel for work.

I am back on Wednesday. I can do Wed-Fri after 6pm UTC. I'll need to check Sat, but I'm more flexible on Sun.

I'm in MST, btw. In Colorado. What about you?

TheHollidayInn:
@goldbergyoni Just following up. Are you free this week?

goldbergyoni:
@TheHollidayInn Sorry for the delay, on a very tight trip now in US and struggle to find time.

My suggestion - let's merge this item as-is now, then there is a couple of enrichment I'd like us to perform as additional PRs. Works for you?

@BrunoScheufler @js-kyle Could you please help with reviewing the formatting/grammar/wording?

p.s. @TheHollidayInn Did you see the credit I gave you in my testing practices guide?

TheHollidayInn:
Works for me! I hope your trip is going well :D

I have your testing practices on my list to read. I just saw it in Node Weekly, congrats!!
 I now see the link to me! Thank you :D :D

goldbergyoni:
@BrunoScheufler Could you please help with format and language review?

goldbergyoni:
@TheHollidayInn I've merged this, big congrats to us and many thanks for the high effort here! 🎇 🚀 🙏 

I believe we can and should take 1/2 level up, I'll pr with some suggestions

PR #379, opened 2019-04-15
6.12 add code example

animir:
Add related code example with `rate-limiter-flexible` package.


js-kyle:
@animir The example here is pretty lengthy - could we trim it down? Making it easy to read and absorb knowledge is our mission, the example here is overloaded. Perhaps a link to the package wiki with more complex examples could be added, with just a basic example provided here?

animir:
@js-kyle Agree, it is too long. I've created another PR https://github.com/i0natan/nodebestpractices/pull/397

Even basic is too long, I added a clue with example of 2 limiters config and a link to complete example.
It looks much better now.

PR #381, opened 2019-04-16
Update README.brazilian-portuguese.md

jjtfsalgado:


i0natan:
@BrunoScheufler Isn't @sagirk handling the Brazilian translation? noted that you assigned to me. I don't mind to handle this, just ensuring there's no mistake

PR #391, opened 2019-04-28
Turkish translation

mperk:


BrunoScheufler:
@mperk if you're fine with me merging it into the translation branch, simply add a confirmation message in this PR thread so I'll know once it's ready 👍

goldbergyoni:
@mperk Looks like you spent here a great amount of time and many content was translated, shall we try to translate at least the 1st page and merge to master?

PR #396, opened 2019-05-04
fix: Typo in centralizedhandling.md files

thealexandrelara:
Fix a small typo that can cause a real application to crash or don't work as expected.

An `err` object is passed to `logger.logError(err)` but the function argument is named `error`.


i0natan:
@thealexandrelara Thanks and welcome aboard 🍺

PR #408, opened 2019-05-24
Fix error handling example

### LINKED ISSUE #405: Possibly confusing error handling example
First of all, thanks for writing this best practices list :slightly_smiling_face:

I've been going through this list as I'm working on an student project using Express.js. I'm trying to follow the best practices when it comes to error handling, but I'm a bit confused by one of the code samples given. In the [Distinguish operational vs programmer errors](https://github.com/i0natan/nodebestpractices/blob/master/sections/errorhandling/operationalvsprogrammererror.md) best practice, it's recommended to not handle programmer errors specifically (other than gracefully restarting the application). However, in [Handle errors centrally. Not within middlewares](https://github.com/i0natan/nodebestpractices/blob/master/sections/errorhandling/centralizedhandling.md), the following example seems to do the opposite:

https://github.com/i0natan/nodebestpractices/blob/7f60837666abc8e023d610b1350175d2b9fc5b08/sections/errorhandling/centralizedhandling.md#L28-L34

If I'm reading it correctly, it's redirecting the error to be handled by middleware (so that a message can be displayed to the user) *only* if it's a programmer error. This seems contradictory to the advice about handling operational errors specifically.

Am I understanding it correctly, or is it a typo in the example?

Calinou:
The error handling middleware should pass all errors to the centralized error handler, not just programmer errors.

This closes #405.

### DIFF
--- a/sections/errorhandling/centralizedhandling.md
+++ b/sections/errorhandling/centralizedhandling.md
@@ -27,24 +27,28 @@ catch (error) {
 
 // Error handling middleware, we delegate the handling to the centralized error handler
 app.use(async (err, req, res, next) => {
-  const isOperationalError = await errorHandler.handleError(err);
-  if (!isOperationalError) {
-    next(err);
-  }
+  await centralizedErrorHandler.handleError(err);
 });
 ```
 
 ### Code example – handling errors within a dedicated object
 
 ```javascript
-module.exports.handler = new errorHandler();
+module.exports.handler = new centralizedErrorHandler();
 
-function errorHandler() {
+function centralizedErrorHandler() {
   this.handleError = async function(err) {
     await logger.logError(err);
     await sendMailToAdminIfCritical;
     await saveInOpsQueueIfCritical;
-    await determineIfOperationalError;
+    const isOperationalError = await determineIfOperationalError(err);
+    if (isOperationalError) {
+      // return an error response to the client
+      return res.status(this.mapErrorToHttp(err));
+    } else {
+      // it's a programmer error, crash to recover the application
+      process.exit(1);
+    }
   };
 }
 ```
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 
 **TL;DR:** Operational errors (e.g. API received an invalid input) refer to known cases where the error impact is fully understood and can be handled thoughtfully. On the other hand, programmer error (e.g. trying to read undefined variable) refers to unknown code failures that dictate to gracefully restart the application
 
-**Otherwise:** You may always restart the application when an error appears, but why let ~5000 online users down because of a minor, predicted, operational error? the opposite is also not ideal – keeping the application up when an unknown issue (programmer error) occurred might lead to an unpredicted behavior. Differentiating the two allows acting tactfully and applying a balanced approach based on the given context
+**Otherwise:** You may always restart the application when an error appears, but why let ~5000 online users down because of a minor, predicted, operational error? The opposite is also not ideal – keeping the application up when an unknown issue (programmer error) occurred might lead to an unpredicted behavior. Differentiating the two allows acting tactfully and applying a balanced approach based on the given context
 
 🔗 [**Read More: operational vs programmer error**](/sections/errorhandling/operationalvsprogrammererror.md)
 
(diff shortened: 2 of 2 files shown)


sections/errorhandling/centralizedhandling.md
```diff
@@ -44,7 +42,12 @@ function errorHandler() {
     await logger.logError(err);
     await sendMailToAdminIfCritical;
     await saveInOpsQueueIfCritical;
-    await determineIfOperationalError;
+
+    const isOperationalError = await determineIfOperationalError;
+    if (!isOperationalError) {
+      // it's a programmer error, crash to recover the application
+      process.exit(1);
```
Calinou: Is this the best way to crash the application? Will process managers such as Forever and PM2 catch it correctly? I haven't tested it yet.

i0natan: @Calinou

Process.exit(1) is the way to do it.

Bottom part should be:
```
await saveInOpsQueueIfCritical;
    const isOperationalError = await determineIfOperationalError(err);
    if (!isOperationalError) {
      // it's a programmer error, crash to recover the application
      process.exit(1);
    }
```

Your first call to determineIfOperationalError seems redundant, no?

i0natan:
@Calinou Looks great to me, except the code comment I shared in my review

Calinou:
I updated the pull request with requested changes.

> Your first call to determineIfOperationalError seems redundant, no?

`determineIfOperationalError` is only being called once :slightly_smiling_face:

i0natan:
> I updated the pull request with the requested changes applied.

Thank you. Two last tiny thoughts below. Just make a call, LMK and I'll merge:

1. After the error middleware passes the error to the centralized hadnler, it calls 'next(err)'. 

I don't it should, it doesn't make sense as the error is already handled. I would just remove that line

2. function errorHandler() {

I would call this 'centralizedErrorHandler'

Calinou:
@i0natan I applied the changes.

> I don't it should, it doesn't make sense as the error is already handled. I would just remove that line

Wouldn't this cause the server to hang in case of operational errors, as it would never pass on to the next middleware? `process.exit()` isn't called by the middleware in case of an operatioanl error.

i0natan:
> I removed that line, but wouldn't this cause the server to hang the request in case of operational errors

First, apologies for the later reply, I just missed that PR 😢 😢 

Good point, I just copied snippets from my main error handler, we should add 2-3 more lines for the 'else' case, whenever it's an operational error, then we should res.status(this.mapErrorToHttp(error())

goldbergyoni:
@Calinou Just a friendly reminder, it looks like we're so close to perfect the code example

Calinou:
@goldbergyoni I just amended the PR, does it look good to you now? :slightly_smiling_face:

goldbergyoni:
> Bump. Is this good to merge now? 😃

I'm truly sorry for missing this (important) fix. The conversation wasn't resolved because my comment with a question were last. You didn't answer but the answer is actually clear, my fauilt anyway.

There are conflicts now because @kevynb added TypeScript examples in this file, I'll take care to update this PR to match the new code examples format. Updates will follow very soon

sections/errorhandling/centralizedhandling.md
```diff
@@ -27,24 +27,28 @@ catch (error) {
...
+function centralizedErrorHandler() {
   this.handleError = async function(err) {
     await logger.logError(err);
     await sendMailToAdminIfCritical;
     await saveInOpsQueueIfCritical;
-    await determineIfOperationalError;
+    const isOperationalError = await determineIfOperationalError(err);
+    if (isOperationalError) {
+      // return an error response to the client
+      return res.status(this.mapErrorToHttp(err));
```
webstacker: I'm not sure this is correct, where is "res" coming from?

Calinou: Now that you mentioned it, I just noticed `res` doesn't come from anywhere…

What would be a good way to solve this? Should it be passed as a parameter to `this.handleError`?

webstacker: I doubt you would want to pass `res` in, as it then ties the function to express. Is the current example not already correct?

```javascript
// Error handling middleware, we delegate the handling to the centralized error handler
app.use(async (err, req, res, next) => {
  const isOperationalError = await errorHandler.handleError(err);
  if (!isOperationalError) {
    next(err);
  }
});
```
The handler performs actions that are common to all errors, then passes off to the caller to handle the specifics e.g. Express would potentially send an appropriate HTTP response, however a non web interface might do something different.

PR #416, opened 2019-06-06
Added replicate the node process for performance section

migramcastillo:
I didn't see anything on the discussion issue to solve, so here is the contribution for the performance section

### DIFF
--- a/README.md
+++ b/README.md
@@ -1034,6 +1034,14 @@ All statements above will return false if used with `===`
 
 <br/><br/><br/>
 
+## ![✔] 7.2. Replicate the Node process
+
+ **TL;DR:** If you really need to use some blocking/heavy compute code in your application, you can’t use any of the recommendation about [non-blocking the event loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/) and you have a managed server with more than a single CPU, you should consider using the Cluster Module native from NodeJS or a library like [PM2](http://pm2.keymetrics.io/docs/usage/cluster-mode/). Cluster mode will allow you to take advantage of your CPU resources to replicate the Single Thread having availability when the Event Loop is blocked.
+
+**Otherwise:** In high peak traffic your users could suffer some latency when the Event Loop is blocked and if you have a multi CPU managed server you will be wasting valuable resources.
+
+<br/><br/><br/>
+
 # Milestones
 
 To maintain this guide and keep it up to date, we are constantly updating and improving the guidelines and best practices with the help of the community. You can follow our [milestones](https://github.com/i0natan/nodebestpractices/milestones) and join the working groups if you want to contribute to this project
--- a/README.md
+++ b/README.md
@@ -1036,7 +1036,7 @@ All statements above will return false if used with `===`
 
 ## ![✔] 7.2. Replicate the Node process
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1034,6 +1034,14 @@ All statements above will return false if used with `===`
 
 <br/><br/><br/>
 
+## ![✔] 7.2. Replicate the Node process
+
+ **TL;DR:** If you really need to use some blocking/heavy compute code in your application, you can’t use any of the recommendation about [non-blocking the event loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop/) and you have a managed server with more than a single CPU, you should consider using the Cluster Module native from NodeJS or a library like [PM2](http://pm2.keymetrics.io/docs/usage/cluster-mode/). Cluster mode will allow you to take advantage of your CPU resources to replicate the Single Thread having availability when the Event Loop is blocked.
```
BrunoScheufler:
```suggestion
 **TL;DR:** If you really need to use some heavy computational work in your application and you can’t make use of the [non-blocking event loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop), and if you have a managed server with more than a single CPU, you should consider using the native Node.js `cluster` module or a library like [PM2](http://pm2.keymetrics.io/docs/usage/cluster-mode). Cluster mode will allow you to take advantage of your CPU resources to replicate the single thread when the Event Loop is blocked.
```

migramcastillo:
> Looks like a great change 💯 I’ve added a suggestion regarding our style guidelines which might be more inline with our existing content 👍

Thanks for the suggestion!

README.md
```diff
@@ -1034,6 +1034,14 @@ All statements above will return false if used with `===`
 
 <br/><br/><br/>
 
+## ![✔] 7.2. Replicate the Node process
+
+ **TL;DR:** If you really need to use some heavy computational work in your application and you can’t make use of the [non-blocking event loop](https://nodejs.org/en/docs/guides/dont-block-the-event-loop), and if you have a managed server with more than a single CPU, you should consider using the native Node.js `cluster` module or a library like [PM2](http://pm2.keymetrics.io/docs/usage/cluster-mode). Cluster mode will allow you to take advantage of your CPU resources to replicate the single thread when the Event Loop is blocked.
```
i0natan: Replication gains better performance even for IO-only work - I tried once to run 1 vs 4 express apps that just serve requests (no CPU work) and saw 20% improvements for 4 processes

I would generalize the advice - if you have more than 1 vCPU, replicate a Node.js process per each

i0natan: We want to modernize this guide and refer to modern platform (k8s, serverless). I would start there and mention that most modern platforms have build-in replication capabilities, in case you deploy to a bare server then consider using some... (cluster, PM2)

README.md
```diff
@@ -1034,6 +1034,14 @@ All statements above will return false if used with `===`
...
+
+**Otherwise:** In high peak traffic your users could suffer some latency when the Event Loop is blocked and if you have a multi CPU managed server you will be wasting valuable resources.
```
i0natan: Sums it up well, consider spicing it up - You pay for 4 cpu and starve 3 of them.

README.md
```diff
@@ -1034,6 +1034,14 @@ All statements above will return false if used with `===`
...
+
```
i0natan: There should be 'Read more' section

PR #417, opened 2019-06-10
All contributors temporary setup

js-kyle:


i0natan:
@js-kyle Will this replace our existing 'flowers'?

js-kyle:
@i0natan Yeah, I think that is what we would want, just need to see how it works. I haven't been able to see the bot working yet, I need to read more about it

i0natan:
> Yeah, I think that is what we would want, just need to see how it works. I haven't been able to see the bot working yet, I need to read more about it

👍 👍

PR #419, opened 2019-06-11
Update code examples in asyncerrorhandling.md

autoboxer:
- makes the Promises example more readable
- adds an async/await example


i0natan:
@autoboxer That's a great contribution! Really makes it more readable & modern (Async await)

We're also having thoughts to add Typescript examples to all sections

p.s. A flower will be assigned by the end of the month :)

PR #424, opened 2019-06-19
Add tags for all-contributors

js-kyle:
Couple more things needed to get the bot working on this test file. After I work out how to prepare the test file how we like it, I'll look at setting it up for the main contributors list (Please approve PR so I can merge)


i0natan:
@js-kyle Good idea to test on a diff file

PR #434, opened 2019-06-30
New BP: testing AAA pattern

i0natan:


README.md
```diff
@@ -452,15 +452,27 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 4.3 Detect code issues with a linter
+## ![✔] 4.3 Structure tests by the AAA pattern
+
+**TL;DR:** Structure your tests with 3 well-separated sections: Arrange, Act & Assert (AAA). The first part includes the test setup, then the execution of the unit under test and finally the assertion phase. Following this structure guarantees that the reader spends no brain CPU on understanding the test plan
+
+
```
oshliaer: There is a single line for the common consistency.

README.md
```diff
@@ -452,15 +452,27 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 4.3 Detect code issues with a linter
```
oshliaer: Is it really important to move down all list (within all translations)?

sections/testingandquality/aaa.md
```diff
@@ -0,0 +1,61 @@
+# Structure tests by the AAA pattern
+
+<br/><br/>
+
+### One Paragraph Explainer
+Our biggest testing challenge is the lack of headspace - we already have the production code keeping us super-busy. For this reason the testing code must stay dead-simple and easy to understand. When reading a test case - it shouldn't feel like reading imperative code (loops, inheritance) rather more like HTML - a declarative experience. To achieve this, keep the AAA convention so the readers mind will parse the test intent effortlessly. There some other similar formats to this pattern, like XUnit 'Setup, Excercise, Verify, Teardown'. These are the three A:
+
+The 1st A - Arrange: All the setup code to bring the system to the scenario the test aims to simulate. This might include instantiating the unit under test constructor, adding DB records, mocking/stubbing on objects and any other preparation code
```
BrunoScheufler: Maybe we could use headings for improved layouts or mark the titles „The 1st A...“ as bold 😄

README.md
i0natan: @oshliaer good catch!

README.md
i0natan: @oshliaer Good question, maybe @BrunoScheufler and @js-kyle can chime-in

I find the order of advice to be important, the first bullets deal with the test structure and the basics. The last bullets deal with high-level stuff like CI. I don't find a reason to mix the topics... on the other end changing the numbers might break links

What are your thoughts?

sections/testingandquality/aaa.md
i0natan: @BrunoScheufler Can you suggest/show something concurrent?

README.md
js-kyle: I really agree that new bullets should be added at the bottom of the section, it's very confusing to shuffle things around when introducing a new best practice

However, yes, I understand there is great value in shuffling things around in the sections, I just thing that should be done in a separate PR so it is clear, and ensure that all the translations are also shuffled at the same time where possible 👍

i0natan: @js-kyle Resonates with me, will separate into a different PR next time

PR #437, opened 2019-07-09
Translate into Korean from 4.6 to 5.10

uronly14me:
This will also delete unnecessary line.


i0natan:
@uronly14me Looks great man. I might miss something, but why is the string "<<<<<<< HEAD" appears in the content multiple time (not the diff view, also when you watch the raw file)?

PR #442, opened 2019-07-20
Fix a typo :haircut_woman: 

oshliaer:
I think there is a typo.

```
, it becomes hairy to match other lines
```
to
```
, it becomes hard to match other lines
```


BrunoScheufler:
I believe this one is actually intended 😄 While marked as informal, "hairy" should probably refer to "risky", "challenging" or "difficult" 👍 Thanks for having a closer look though 🙌

oshliaer:
@BrunoScheufler ,

Thank. It is valuable to me! :joy:

i0natan:
@oshliaer This is just great that as you translate you spot opportunities to improve, we'd love to see more

oshliaer:
Usually, after the translation, I once again read the original and the finished translation. I did not learn English anywhere (only by myself) and some things are completely incomprehensible to me =)

PR #448, opened 2019-08-02
Russian translation. Removed EN fragment

nicotrygub:
Hi everyone. This PR removes original EN paragraph that was translated by @oshliaer


oshliaer:
It seems there should be a mistake somewhere else. Because the number of lines must match.

BrunoScheufler:
@nicotrygub thanks for noticing and fixing this!

PR #465, opened 2019-08-20
Fix absolute link redirecting

AbdelrahmanHafez:
Fixes an issue where the link would redirect to another page (due to the i0natan bit, I guess).


AbdelrahmanHafez:
Should this refer to the bullet point instead of the parent?

oshliaer:
Please look at the 468th line. It seems the link needs to change too.

AbdelrahmanHafez:
@oshliaer Good catch! Just changed it.

PR #575, opened 2019-10-10
Added a bullet point on validating env vars at start-up

### LINKED ISSUE #561: Recommend asserting the presence of env variables required for running the app
Sometimes environment variables are used by the code, which is not exercised at startup, e.g. one of the routes of the server.
If the variable is not set, the application will fail at run-time, potentially in production, affecting the end-users.

The guide already recommends minimizing the number of env vars, so I feel like the next step would be recommending on asserting the presence of all required env vars at start-up.

```js
const assert = require('assert')

assert(
  ['dev','prod'].includes(proccess.env.NODE_ENV),
  'NODE_ENV env variable is required for application to run'
)
```

This way we can make sure the app will fail asap if required env vars are not present.

What do you think?

halfzebra:
Closes #561

### DIFF
--- a/sections/projectstructre/configguide.md
+++ b/sections/projectstructre/configguide.md
@@ -14,6 +14,8 @@ When dealing with configuration data, many things can just annoy and slow down:
 
 4. some advanced configuration scenarios demand to inject configuration values via command line (vargs) or sync configuration info via a centralized cache like Redis so multiple servers will use the same configuration data.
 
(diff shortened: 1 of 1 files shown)


halfzebra:
Thanks for merging the PR!

goldbergyoni:
@halfzebra Thank you 🚀 🙏 

Our bot should add your name to the contributors list, if not - please ping

PR #582, opened 2019-10-26
ADD - Error Handling for individual 'awaited' promise using .catch

bhatnagardivyanshu:
The async/await errors can also be handled using a '.catch' at the end of the awaited function.

### DIFF
--- a/sections/errorhandling/asyncerrorhandling.md
+++ b/sections/errorhandling/asyncerrorhandling.md
@@ -31,6 +31,26 @@ async function executeAsyncTask () {
     await alwaysExecuteThisFunction();
   }
 }
+
+async function executeAsyncTask () {
+  const valueA = await functionA();
+  try {
+    const valueA = await functionA();
+    const valueB = await functionB(valueA).catch(err => {
+      throw new Error('Something went wrong while executing function B'); // stops the further execution of the try block
+    });
+
+    console.log('This will never run if functionB returns a rejected Promise');
+    const valueC = await functionC(valueB);
+    return await functionD(valueC);
+  }
+  catch(err) {
+    logger.error(err);
+  } finally {
+    await alwaysExecuteThisFunction();
+  }
+}
+
 ```
 
 ### Anti pattern code example – callback style error handling


goldbergyoni:
@bhatnagardivyanshu Welcome abroad! 🍺 

What is the point we're trying to make here? another syntax? does it imply any best practice or a good way of doing things?

bhatnagardivyanshu:
Hey, thank you for the warm welcome.
Yeah that's something not known by many developers that we can simply catch errors for the 'awaited' calls and hence eliminate the use of the try catch. It's just another way of handling the errors. Thought it'd be really helpful for better understanding and writing a more clean code.

bhatnagardivyanshu:
@goldbergyoni Hey, just wanted to remind you about this.

goldbergyoni:
@bhatnagardivyanshu Oh sorry, I wasn't clear, this is a reminder bot that automatically notify. I didn't mean that you'll remind me but thanks for being kind and doing this :)

Processing this in the next 2 days, speak soon

bhatnagardivyanshu:
Oh, never mind.

Looking forward to it. Thanks.
And also, thank you for creating such a powerful resource of information.

goldbergyoni:
@bhatnagardivyanshu Had some thoughts on this, I think that we shouldn't emphasize this possibility. Not anything that is possible is also right. Why do I believe that it should be avoided? using async/await is a cleaner pattern and classic try-catch will catch both rejections and also sync code errors. Promises catch will only catch errors that originated from a promise - should we have some function that starts with non-promise code and then invokes promises, all the erorrs of the first part won't get caught

For example:
```
try{
const result = await someIOFunc();
const calculation = someNonIOFunc();
}
catch(){
//will catch all type of errors
}

```

Does this make sense?

bhatnagardivyanshu:
Hey @goldbergyoni, I understand and agree with your point. However, one case where I find this extremely useful is where let's say, in the example you have given, I want to run the **someNonIOFunc()** even if the **someIOFunc()** fails. 

To support this requirement, I'll have to either do it finally or find some another work-around. Considering there could be a number of other functionalities which I always want to execute even if the **someIOFunc()** fails.

Let's say

`let someParameter = 1;`
`const result = await someIOFunc().catch(() => someParameter = -1);`
` const calculation = someNonIOFunc(someParameter);`

I think in such conditions it can be really useful. What say?

goldbergyoni:
> I think in such conditions it can be really useful. What say?

I'd say it's absolutely a valid point and looks cleaner than try-catch. On the contrary side, it's a corner case + I find it useful to stick to one pattern (i.e. try-catch). Given all of this, our goal here is to come with a simple piece of advice and I believe that try-catch is the answer to most scenarios and we should stick to it

Thoughts?

bhatnagardivyanshu:
I just thought people should know about it. I understand that try catch is
cleaner and might be a better approach but my point was to let people know
about it. Maybe they find a more useful use case for that. That's the major
motivation behind it.

I am okay with whatever decision you make though. :)

goldbergyoni:
Bottom line: Great language/syntax advice, less relevant as a best practice in this context. Closing and wish to see @bhatnagardivyanshu with more ideas again

Issue #579, opened 2019-10-16
TypeScript examples


kevynb:
Hello,

I saw in the Readme that you're looking for Typescript examples. I think I'm up to the task.

What do you need? Is it just another code block under existing examples using Typescript types?

goldbergyoni:
@kevynb Welcome and sorry for the late reply, I was traveling recently and only now found some free time

Yes, your proposal sounds great. I would create one example, to see how the format looks like, and then discuss this with the team here @BrunoScheufler and @js-kyle and once we love the format we can go to the distance

Just an idea - we can maybe use HTML detail tag to place two examples (JS & TS), so the user will see a title with the code explanation (e.g. throwing an error properly) and then expand it to see JS & TS examples

goldbergyoni:
### Code example: deciding whether to crash

<details><summary><b>JavaScript</b></summary>

```javascript
// Assuming developers mark known operational errors with error.isOperational=true, read best practice #3
process.on('uncaughtException', function(error) {
  errorManagement.handler.handleError(error);
  if(!errorManagement.handler.isTrustedError(error))
  process.exit(1)
});

// centralized error handler encapsulates error-handling related logic
function errorHandler() {
  this.handleError = function (error) {
    return logger.logError(err)
      .then(sendMailToAdminIfCritical)
      .then(saveInOpsQueueIfCritical)
      .then(determineIfOperationalError);
  }

  this.isTrustedError = function (error) {
    return error.isOperational;
  }
}
```
</details>

<br/>
<details><summary><b>TypeScript</b></summary>

```javascript
// Assuming developers mark known operational errors with error.isOperational=true, read best practice #3
process.on('uncaughtException', function(error) {
  errorManagement.handler.handleError(error);
  if(!errorManagement.handler.isTrustedError(error))
  process.exit(1)
});

// centralized error handler encapsulates error-handling related logic
function errorHandler() {
  this.handleError = function (error) {
    return logger.logError(err)
      .then(sendMailToAdminIfCritical)
      .then(saveInOpsQueueIfCritical)
      .then(determineIfOperationalError);
  }

  this.isTrustedError = function (error) {
    return error.isOperational;
  }
}
```
</details>

kevynb:
Yeah, that looks good to me! 
I'll start converting some code while we wait for confirmation from the others.

PR #584, opened 2019-10-27
Add typescript examples for error handling and project structure

kevynb:
This PR adds typescript examples to the infrastructure and error handling sections.

I've formatted the code as proposed in #579 

If that style suits everyone I will add examples to the other sections too.

I've had to add `<br/>` to get linebreaks to appear in the `<pre>` block on github.

### DIFF
--- a/sections/errorhandling/asyncerrorhandling.md
+++ b/sections/errorhandling/asyncerrorhandling.md
@@ -49,7 +49,7 @@ getData(someParameter, function(err, result) {
                 getMoreData(b, function(c) {
                     getMoreData(d, function(e) {
                         if(err !== null ) {
-                            // you get the idea? 
+                            // you get the idea?
                         }
                     })
                 });
@@ -73,7 +73,7 @@ getData(someParameter, function(err: Error | null, resultA: ResultA) {
         getMoreData(resultB, function(resultC: ResultC) {
           getMoreData(resultC, function(err: Error | null, d: ResultD) {
             if(err !== null) {
-              // you get the idea? 
+              // you get the idea?
             }
           })
         });
--- a/sections/errorhandling/asyncerrorhandling.md
+++ b/sections/errorhandling/asyncerrorhandling.md
@@ -8,12 +8,13 @@ Callbacks don’t scale well since most programmers are not familiar with them. 
 
 ```javascript
 return functionA()
-  .then((valueA) => functionB(valueA))
-  .then((valueB) => functionC(valueB))
-  .then((valueC) => functionD(valueC))
+  .then(functionB)
+  .then(functionC)
+  .then(functionD)
   .catch((err) => logger.error(err))
-  .then(alwaysExecuteThisFunction())
+  .then(alwaysExecuteThisFunction)
 ```
+
 
 ### Code Example - using async/await to catch errors
 
@@ -25,7 +26,7 @@ async function executeAsyncTask () {
     const valueC = await functionC(valueB);
     return await functionD(valueC);
   }
-  catch(err) {
+  catch (err) {
     logger.error(err);
   } finally {
     await alwaysExecuteThisFunction();
@@ -34,6 +35,9 @@ async function executeAsyncTask () {
 ```
 
 ### Anti pattern code example – callback style error handling
+
+<details>
+<summary><strong>Javascript</strong></summary>
 
 ```javascript
 getData(someParameter, function(err, result) {
@@ -54,6 +58,31 @@ getData(someParameter, function(err, result) {
     }
 });
 ```
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+
+```typescript
+getData(someParameter, function(err: Error | null, resultA: ResultA) {
+  if(err !== null) {
+    // do something like calling the given callback function and pass the error
+    getMoreData(resultA, function(err: Error | null, resultB: ResultB) {
+      if(err !== null) {
+        // do something like calling the given callback function and pass the error
+        getMoreData(resultB, function(resultC: ResultC) {
+          getMoreData(resultC, function(err: Error | null, d: ResultD) {
+            if(err !== null) {
+              // you get the idea? 
+            }
+          })
+        });
+      }
+    });
+  }
+});
+```
+</details>
 
 ### Blog Quote: "We have a problem with promises"
 
--- a/sections/errorhandling/catchunhandledpromiserejection.md
+++ b/sections/errorhandling/catchunhandledpromiserejection.md
@@ -4,7 +4,7 @@
 
 ### One Paragraph Explainer
 
-Typically, most of modern Node.js/Express application code runs within promises – whether within the .then handler, a function callback or in a catch block. Surprisingly, unless a developer remembered to add a .catch clause, errors thrown at these places are not handled by the uncaughtException event-handler and disappear.  Recent versions of Node added a warning message when an unhandled rejection pops, though this might help to notice when things go wrong but it's obviously not a proper error handling method. The straightforward solution is to never forget adding .catch clauses within each promise chain call and redirect to a centralized error handler. However, building your error handling strategy only on developer’s discipline is somewhat fragile. Consequently, it’s highly recommended using a graceful fallback and subscribe to `process.on(‘unhandledRejection’, callback)` – this will ensure that any promise error, if not handled locally, will get its treatment.
+Typically, most of modern Node.js/Express application code runs within promises – whether within the .then handler, a function callback or in a catch block. Surprisingly, unless a developer remembered to add a .catch clause, errors thrown at these places are not handled by the uncaughtException event-handler and disappear.  Recent versions of Node added a warning message when an unhandled rejection pops, though this might help to notice when things go wrong but it's obviously not a proper error handling method. The straightforward solution is to never forget adding .catch clauses within each promise chain call and redirect to a centralized error handler. However, building your error handling strategy only on developer’s discipline is somewhat fragile. Consequently, it’s highly recommended using a graceful fallback and subscribe to `process.on('unhandledRejection', callback)` – this will ensure that any promise error, if not handled locally, will get its treatment.
 
 <br/><br/>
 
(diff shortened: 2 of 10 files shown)


sections/errorhandling/failfast.md
```diff
@@ -10,35 +10,72 @@ Defensive programming is an approach to improve software and source code, in ter
...
+<pre lang="typescript">
+// if the discount is positive let's then redirect the user to print his discount coupons
+function redirectToPrintDiscount(httpResponse: Response, member: Member, discount: number) {
+  if (discount != 0) {
+    httpResponse.redirect(`/discountPrintView/${member.id}`);
+  }
+}
+<br/>
+redirectToPrintDiscount(httpResponse, someMember, -12);
+// We passed a negative parameter discount, why the heck was the user redirected to the discount screen?
```
kevynb: I altered the original comment here, since we cannot forget to pass the parameter in typescript

sections/errorhandling/operationalvsprogrammererror.md
```diff
@@ -21,10 +23,37 @@ class AppError {
...
+<pre lang="typescript">
+// some centralized error factory (see other examples at the bullet "Use only the built-in Error object")
+export class AppError extends Error {
+  public readonly commonType: string;
+  public readonly isOperational: boolean;
+<br/>
+  constructor(commonType: string, description: string, isOperational: boolean) {
+    super(description);
+<br/>
+    Object.setPrototypeOf(this, new.target.prototype); // restore prototype chain
```
kevynb: There is an explanation for this line in `useonlythebuiltinerror.md`. Should we state it again here?

sections/errorhandling/shuttingtheprocess.md
```diff
@@ -22,12 +24,56 @@ function errorHandler() {
...
+class ErrorHandler {
+  public async handleError(err: Error): Promise<void> {
+    await logger.logError(err);
+    await sendMailToAdminIfCritical();
+    await saveInOpsQueueIfCritical();
+    await determineIfOperationalError();
+  };
+<br/>
+  public isTrustedError(error: Error) {
+    if (error instanceof AppError) {
```
kevynb: Given that we cannot have `isOperational` on `Error` type, I added this check.

sections/errorhandling/testingerrorflows.md
```diff
@@ -6,20 +6,37 @@ Testing ‘happy’ paths is no better than testing failures. Good testing code
 
 ### Code example: ensuring the right exception is thrown using Mocha & Chai
 
-```javascript
+<details>
+<summary><strong>Javascript</strong></summary>
+<pre lang="javascript">
 describe("Facebook chat", () => {
   it("Notifies on new chat message", () => {
     var chatService = new chatService();
```
kevynb: We could switch the `var` for a `const` and only use one example here instead of two?

kevynb: Also, but that might not be the place to discuss this, mocha state that we should avoid using arrow functions for tests: https://mochajs.org/#arrow-functions. So should we switch to `function` instead? Or maybe that's another PR?

sections/errorhandling/testingerrorflows.md
```diff
@@ -35,4 +52,28 @@ it("Creates new Facebook group", function (done) {
...
 });
-```
+</pre>
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+<pre lang="typescript">
+it("Creates new Facebook group", async function (done) {
+  let invalidGroupInfo = {};
```
kevynb: Same thing here, I think that by "modernizing" the javascript code, we could remove the need for a Typescript example. WDYT?

sections/errorhandling/usematurelogger.md
```diff
@@ -9,38 +9,78 @@ We all love console.log but obviously, a reputable and persistent logger like [W
...
 logger.log('info', 'Test Log Message with some parameter %s', 'some parameter', { anything: 'This is metadata' });
+</pre>
+</details>
 
-```
+<details>
+<summary><strong>Typescript</strong></summary>
+<pre lang="typescript">
+// your centralized logger object
+const logger = new winston.Logger({
```
kevynb: Also unify code examples here?

sections/projectstructre/separateexpress.md
```diff
@@ -8,46 +8,58 @@ The latest Express generator comes with a great practice that is worth to keep -
...
+<pre lang="javascript">
 var app = require('../app');
 var http = require('http');
-
-/**
- * Get port from environment and store in Express.
- */
-
+<br/>
+// Get port from environment and store in Express.
```
kevynb: I had to change the comment style to avoid weird display bugs. `*` were interpreted as markdown bullet points

sections/errorhandling/testingerrorflows.md
BrunoScheufler: I think we _could_ definitely swap out the arrow functions in places where we have to use the Mocha context as stated in the docs, but for the rest it should definitely be fine (for now). If you'd like to swap those out this would be awesome though!

sections/errorhandling/shuttingtheprocess.md
BrunoScheufler: That's definitely the right path here 😄

sections/errorhandling/testingerrorflows.md
BrunoScheufler: Sounds great, we should maybe do this in a separate PR 👍

sections/errorhandling/usematurelogger.md
BrunoScheufler: Which parts would you unify here? The two code examples into one? I believe these sections were intended in the beginning, but I'm not completely sure here, maybe ask @goldbergyoni about this

kevynb:
Just to understand what you suggest, you would wrap the ‘’’ in a <summary> block? Or is it something else?

BrunoScheufler:
I thought of mostly writing the code we include in our Markdown pages using GitHub's supported synax which starts with three `s (```) followed by the language so writing code like
<pre>
```typescript
console.log('Hello world');
```
</pre>

Should do the trick, right? And if the collapsible section doesn't work, maybe add divs around it? Just thinking how we could get rid of the <br> tags altogether

sections/errorhandling/usematurelogger.md
kevynb: Yes, I would only use one example for both languages since there's not much difference

kevynb:
Ok, I found out a way to get `<details>` and \``` working together. They just need to be separated by a blank line. PR updated

sections/errorhandling/asyncerrorhandling.md
```diff
@@ -54,6 +98,31 @@ getData(someParameter, function(err, result) {
...
+getData(someParameter, function(err: Error | null, resultA: ResultA) {
+  if(err !== null) {
+    // do something like calling the given callback function and pass the error
+    getMoreData(resultA, function(err: Error | null, resultB: ResultB) {
+      if(err !== null) {
+        // do something like calling the given callback function and pass the error
+        getMoreData(resultB, function(resultC: ResultC) {
+          getMoreData(resultC, function(err: Error | null, d: ResultD) {
+            if(err !== null) {
+              // you get the idea? 
```
BrunoScheufler: Just a question about the trailing symbol, what should be displayed here? On my device, I can only see a box 😄

sections/errorhandling/catchunhandledpromiserejection.md
```diff
@@ -4,38 +4,79 @@
...
+```
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
 
+```typescript
+DAL.getUserById(1).then((johnSnow: User) => {
+  // this error will just vanish
+  if(johnSnow.isAlive == false)
```
BrunoScheufler: We should probably stick to the convention of using triple-equals for conditional checks `===`

sections/errorhandling/asyncerrorhandling.md
kevynb: Same thing here, I've got an arrow when editing with my IDE but not in my browser. It was in the js example, just copy-pasted the thing. Should we remove it?

sections/errorhandling/failfast.md
```diff
@@ -37,8 +62,24 @@ function redirectToPrintDiscount(httpResponse, member, discount) {
...
+```
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+
+```typescript
+// if the discount is positive let's then redirect the user to print his discount coupons
+function redirectToPrintDiscount(httpResponse: Response, member: Member, discount: number) {
+  if (discount != 0) {
```
BrunoScheufler: I'm not sure about this example both in JavaScript and TypeScript since it's rather about language features than validation in this case, right? We don't have to remove it completely but in general the solution would rather be to use a different equality operator (`!==`) in this case to avoid implicit type conversion from the undefined value coming from the unsupplied discount and rather only check on the numerical value

kevynb: That's true, maybe the example does not perfectly suit the point we want to illustrate here. 

What come to my mind when I think about this problem is what often happens when we receive a big json object with one missing value and forgot to check it. It goes all the way to the DB to get a constraint error (or worse, if there's not constraint). That could be avoided with schema validation. But that's harder to illustrate. Any other idea?

BrunoScheufler: Yeah that's definitely the right use case for validation, I think the "no validation yields bugs" example can be visualized with various cases, for example sending an email with some helper function and receiving an invalid value causing a crash, these things will always need to be validated, I guess.

kevynb: Thinking a bit more about this, shouldn't this be another PR?
It's more about the existing example not accurately describing the problem than the translation of it in Typescript. I feel like that should be discussed in a separate issue.

goldbergyoni: @kevynb Yes, sounds like another PR but it's a good point that you and @BrunoScheufler raised

sections/errorhandling/failfast.md
goldbergyoni: @kevynb Yes, good point

sections/errorhandling/testingerrorflows.md
goldbergyoni: @BrunoScheufler @kevynb In that case, we won't include sometimes TS examples?

kevynb: Yes, in those case I think we could only display a JavaScript example. I think in some simple case having a typescript code example does not add value.

sections/errorhandling/asyncerrorhandling.md
```diff
@@ -54,6 +98,31 @@ getData(someParameter, function(err, result) {
     }
 });
 ```
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+
+```typescript
+getData(someParameter, function(err: Error | null, resultA: ResultA) {
```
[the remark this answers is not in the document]
kevynb: I don’t know, in was asking myself the same question.
I guess that should be a group decision.

For my part I tend to use arrow function every time I have to write an anonymous function.

sections/errorhandling/asyncerrorhandling.md
```diff
@@ -54,6 +98,31 @@ getData(someParameter, function(err, result) {
...
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+
+```typescript
+getData(someParameter, function(err: Error | null, resultA: ResultA) {
+  if(err !== null) {
+    // do something like calling the given callback function and pass the error
+    getMoreData(resultA, function(err: Error | null, resultB: ResultB) {
```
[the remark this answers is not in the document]
kevynb: It’s an ‘or’. So that means ‘Error or null’

kevynb:
I think I'll find the time to do it by myself.
I'll try to do one PR that "modernize" the js and then I'll build on it for the ts.

sections/errorhandling/asyncerrorhandling.md
```diff
@@ -32,9 +52,33 @@ async function executeAsyncTask () {
   }
 }
 ```
+</details>
+
+<details>
+<summary><strong>Typescript</strong></summary>
+
+```typescript
+async function executeAsyncTask () {
```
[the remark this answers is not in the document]
kevynb: I didn't type the variables because the TS compiler should be able to infer the return types of the functions. However I'll add a return type to `executeAsyncTask`, good catch!

kevynb:
I might also squash the commits before we merge that one.

sections/errorhandling/testingerrorflows.md
goldbergyoni: @kevynb It might look inconsistent and fool readers to believe that it wasn't done yet. Some may contribute it

Maybe include a title 'TypeScript - same as JS'?

goldbergyoni:
@kevynb Let me know then when to merge

sections/errorhandling/testingerrorflows.md
kevynb: I'm not sure it's a good idea to make the doc more verbose just to prevent a contributor from making a contrib. Maybe we should state somewhere that Typescript examples need to be added only where they add a value?

kevynb:
It should be good to go :)

sections/errorhandling/testingerrorflows.md
goldbergyoni: Consistency is an important part of readability - people absorb the format and tends to look for the same. Anyway, that's not critical for now, let's go with your thoughts

goldbergyoni:
@kevynb That's an epic commit 🚀 ❤️

goldbergyoni:
https://twitter.com/nodepractices/status/1196146346807451648?s=20

PR #591, opened 2019-11-13
Further summarized the TLDR 

haider-ilahi:


goldbergyoni:
@haider-ilahi Welcome aboard and thanks for the care to make things better!

In this case, my gut feeling is keeping the origin test which is indeed longer than usual - why? when discussing architecture things tend to be vague and abstract so it was important to me to include some detail and make tangile

Does this make sense?

goldbergyoni:
@haider-ilahi I'm closing this. The overall curation intent is right and thoughtful, only in this specific case I find it important to be verbose. Thanks for respecting my judgement. Would really be glad to see other ideas coming from you!

PR #593, opened 2019-11-21
refactor: replace deprecated __proto__ in built-in Error example

tiendq:
[`__proto__`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/proto) is deprecated and shouldn't be used to set object prototype.


BrunoScheufler:
@all-contributors-bot please add @tiendq for content

PR #616, opened 2020-01-14
Update nativeoverutil.md

germangamboa95:
The reason for file change is to correct the statement that the original author created to mean what was intended.

### DIFF
--- a/sections/performance/nativeoverutil.md
+++ b/sections/performance/nativeoverutil.md
@@ -4,8 +4,7 @@
 <br/><br/>
 
 ### One Paragraph Explainer
-
-Sometimes, using native methods is better than requiring `lodash` or `underscore` because it will not lead to performance loss or take up more space than needed.
+Sometimes, using native methods is better than requiring _lodash_ or _underscore_ because those libraries can lead to performance loss or take up more space than needed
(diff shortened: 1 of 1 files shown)


sections/performance/nativeoverutil.md
```diff
@@ -5,7 +5,7 @@
 
 ### One Paragraph Explainer
 
-Sometimes, using native methods is better than requiring `lodash` or `underscore` because it will not lead in a performance boost and use more space than necessary.
+Sometimes, using native methods is better than requiring `lodash` or `underscore` because it will not lead to performance loss or take up more space than needed.
```
kevynbsomfy: Hi @germangamboa95 
Couldn't we rephrase this sentence to avoid the negation? I think the `it will not lead` makes the sentence a bit harder to understand.

Something like `Sometimes, using native methods is better than requiring _lodash_ or _underscore_ because those libraries can lead to performance loss or take up more space than needed` ?
I'm not set on that exact sentence, just clarifying what I mean :)

kevynb: Hi @germangamboa95 
Couldn't we rephrase this sentence to avoid the negation? I think the `it will not lead` makes the sentence a bit harder to understand.

Something like `Sometimes, using native methods is better than requiring _lodash_ or _underscore_ because those libraries can lead to performance loss or take up more space than needed` ?
I'm not set on that exact sentence, just clarifying what I mean :)

germangamboa95: I agree on removing the double negative. Your sentence is good. I'll update the PR.

goldbergyoni:
@germangamboa95 Welcome and thanks

@kevynb You have permissions now to merge yourself

@all-contributors please add @germangamboa95  for content

PR #627, opened 2020-02-14
docs: Add a section about using npm ci

kevynb:
Hi there,

I added a section about the use of npm ci.
I thought it would be better to put it right after the bullet point about locking dependencies.

Let me know what you think of it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 2. [Error Handling Practices (11) ](#2-error-handling-practices)
 3. [Code Style Practices (12) ](#3-code-style-practices)
 4. [Testing And Overall Quality Practices (12) ](#4-testing-and-overall-quality-practices)
-5. [Going To Production Practices (18) ](#5-going-to-production-practices)
+5. [Going To Production Practices (19) ](#5-going-to-production-practices)
 6. [Security Practices (25)](#6-security-best-practices)
 7. [Performance Practices (2) (Work In Progress️ ✍️)](#7-draft-performance-best-practices)
 
@@ -719,6 +719,16 @@ All statements above will return false if used with `===`
 **Otherwise:** Application handling log routing === hard to scale, loss of logs, poor separation of concerns
 
 🔗 [**Read More: Log Routing**](/sections/production/logrouting.md)
+
+<br/><br/>
+
+## ![✔] 5.19. Install your packages with `npm ci`
+
+**TL;DR:** You have to be sure that production code uses the exact version of the packages you have tested it with. Run `npm ci` to do a clean install of your dependencies matching package.json and package-lock.json.
+
+**Otherwise:** QA will thoroughly test the code and approve a version that will behave differently in production. Even worse, different servers in the same production cluster might run different code
+
+🔗 [**Read More: Use npm ci**](/sections/production/installpackageswithnpmci.md)
 
 <br/><br/><br/>
 
--- a/sections/production/installpackageswithnpmci.md
+++ b/sections/production/installpackageswithnpmci.md
@@ -1,0 +1,30 @@
+# Install packages with npm ci in production
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+You locked your dependencies following [**Lock dependencies**](/sections/production/lockdependencies.md) but you now need to make sure those exact package versions are used in production.
+
+Using `npm ci` to install packages will do exactly that and more.
+* It will fail if your `package.json` and your `package-lock.json` do not match (they should) or if you don't have a lock file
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -594,7 +604,7 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 5.6. Utilize all CPU cores
+## ![✔] 5.7. Utilize all CPU cores
```
js-kyle: It's my preference to add new bullets to the bottom - otherwise bullet numbers in translation files no longer match up. Unsure if we agree on that though, so don't worry about changing it

sections/production/installpackageswithnpmci.md
```diff
@@ -0,0 +1,30 @@
+# Install packages with npm ci in production
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+You locked your dependencies following [**Lock dependencies**](/sections/production/lockdependencies.md) but you now need to make sure those exact package versions are used in production.
+
+Using `npm ci` to install packages will do exactly that and more.
+* It will only install packages in the `dependencies` section of your `package.json`. `devDependencies` are ignored
```
js-kyle: `npm ci` does install devDependencies?

sections/production/installpackageswithnpmci.md
```diff
@@ -0,0 +1,30 @@
...
+* It will fail if your `package.json` and your `package-lock.json` do not match (they should) or if you don't have a lock file
+* If a `node_modules` folder is present it will automatically remove it before installing
+* It is faster! Nearly twice as fast according to the release blog post (see link below)
```
js-kyle: Could be nice to also link it here

README.md
kevynb: I wasn't sure either, that's why it in a separate commit :)
Do you want to wait for others to chime in?

sections/production/installpackageswithnpmci.md
kevynb: No. If you're asking, I guess the point is not clear enough! I'll try to rewrite that differently

sections/production/installpackageswithnpmci.md
kevynb: So you would have the link in both sections? Or just here?

README.md
```diff
@@ -584,7 +584,17 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 5.5. Guard process uptime using the right tool
+## ![✔] 5.5. Install your packages with `npm ci`
+
+**TL;DR:** You have to be sure that production code uses the exact version of the packages you have tested it with. Run `npm ci` to do a clean install of your dependencies matching package.json and package-lock.json.
```
goldbergyoni: Very concise and drives the point home

sections/production/installpackageswithnpmci.md
```diff
@@ -0,0 +1,30 @@
...
+
+As per the first point, it's good to note that `devDependencies` have nothing to do on a production server (hence the name).
```
goldbergyoni: Why they have nothing to do on a production? answer the 'so what' question, why is this significant: Less packages means minimized attack surface

sections/production/installpackageswithnpmci.md
```diff
@@ -0,0 +1,30 @@
...
+It also reduces the size of the bundle of code you are shipping (for example inside a docker image).
+
```
goldbergyoni: Makes it even more tangible, show an example of when this can happen:
1. Someone changes manually package.json, not through a cli command but rather direct package.json editing. Then a gap between package.json & .lock is presented
2. An attacker changes only the lock to refer to a malicious package/version (becauase no one review these huge files), if that malicious dep won't exist in package.json the command will fail (I think)

sections/production/installpackageswithnpmci.md
js-kyle: Sorry I meant to question the line, as `npm ci` does install devDependencies as well as dependencies.

It operates in the same way as npm install where you need to pass either the NODE_ENV variable as `production` or the `--production` cli flag

kevynb: Damn that's right, that's a really good catch. Thank you! I'll reword that part

sections/production/installpackageswithnpmci.md
kevynb: I checked, the second point doesn't hold (and yet I hoped it would 😞 ) You have to check that by manually looking through the dependency list.
I also found https://github.com/lirantal/lockfile-lint but it seems that if the malicious package is hosted on npm.js you're done.

sections/production/installpackageswithnpmci.md
kevynb: I'll create a new BP about using the --production flag in a new PR

kevynb:
PR updated :)

PR #633, opened 2020-02-26
Tests middleware isolation

goldbergyoni:


sections/testingandquality/test-middlewares.md
```diff
@@ -0,0 +1,30 @@
...
+
+<br/><br/>
+
+### Code example: Testing middleware in isolation
+
+```javascript
+//the middleware we want to test
+const unitUnderTest = require("./middleware");
+const httpMocks = require("node-mocks-http");
+//Jest syntax, equivelant to describe() & it() in Mocha
```
kevynb: Little typo: `equivalent`

kevynb: Should we use `jest` syntax here given that we are using mocha syntax everywhere else?

sections/testingandquality/test-middlewares.md
```diff
@@ -0,0 +1,30 @@
+# Test your middlewares in isolation
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+Many avoid Middleware testing because they represent a small portion of the system and require a live Express server. Both reasons are wrong — Middlewares are small but affect all or most of the requests and can be tested easily as pure functions that get {req,res} JS objects. To test a middleware function one should just invoke it and spy ([using Sinon for example](https://www.npmjs.com/package/sinon)) on the interaction with the {req,res} objects to ensure the function performed the right action. The library [node-mock-http](https://www.npmjs.com/package/node-mocks-http) takes it even further and factors the {req,res} objects along with spying on their behavior. For example, it can assert whether the http status that was set on the res object matches the expectation (See example below)
```
kevynb: I think that we could put {req,res} between backticks since it's code: `{req,res}`. It might be more explicit.

README.md
```diff
@@ -392,18 +391,18 @@ module.exports.SMSNumberResolver = require('./SMSNumberResolver/SMSNumberResolve
 ### 3.10 Code example
 
 ```javascript
-'' == '0'           // false
-0 == ''             // true
-0 == '0'            // true
+"" == "0"; // false
```
kevynb: Do we have a convention around quotes or double quotes for code examples?

sections/testingandquality/test-middlewares.md
goldbergyoni: > Should we use `jest` syntax here given that we are using mocha syntax everywhere else?

I would suggest that Yes, why not. Testing shouldn't be locked down to one tool. We can add a comment '// Using Jest, Mocha syntax is almost identical

sections/testingandquality/test-middlewares.md
goldbergyoni: Good idea! Done

kevynb:
Well, I'm used to a prettier configured with single quotes. I have a preference for single quotes but I'm not gonna argue for or against them. I would say let's just decide of a common convention and use it all over the repo. Because it seemed to me last time with a quick skim that we seemed to have more single quotes than double quotes for js code (but my search might have been biased or incomplete 🤔 ).

PR #644, opened 2020-03-01
Updated the code example

chandiran-dmc:
The code example for do and avoid was the same,


kevynb:
Good catch, it seems it's a mistake from the previous MR. Thanks for taking the time to create a PR!

kevynb:
@all-contributors please add @chandiran-dmc for content

PR #652, opened 2020-03-19
fix '.html' out of link

kennethfreitas:


kevynb:
Good catch! Thanks @kennethfreitas

kevynb:
@all-contributors please add @kennethfreitas  for content

PR #681, opened 2020-05-12
added require for the testing package

keyrwinz:


kevynb:
Hello @keyrwinz, thanks for taking the time to contribute to our repository. However, I don't really see the purpose of the code you added. Given that this example is about a "production" node server, why would one use supertest here?

Could you please elaborate?

keyrwinz:
> Hello @keyrwinz, thanks for taking the time to contribute to our repository. However, I don't really see the purpose of the code you added. Given that this example is about a "production" node server, why would one use supertest here?
> 
> Could you please elaborate?

Hi, It's for the example in testing API in-process, it stated there it uses supertest package but never declared in the code.

kevynb:
Ah yes! I missed that in the diff. I re-read the whole file and you’re indeed right, we’re talking about supertest but not using it :)
That’s a good catch!

Can you update the typescript example to use an import instead of a require?

keyrwinz:
> Ah yes! I missed that in the diff. I re-read the whole file and you’re indeed right, we’re talking about supertest but not using it :)
> That’s a good catch!
> 
> Can you update the typescript example to use an import instead of a require?

resolved in 227e00d

kevynb:
@all-contributors please add @keyrwinz for content

PR #687, opened 2020-05-15
Add suggestion to use OpenAPI/Swagger-spec based validation

kibertoad:
Full disclosure - I am a contributor to the second mentioned library. If you feel that this constitutes a conflict of interest and prefer reference to said library to be removed, I would be happy to do so (but it provides support for `koa` and `fastify`, while the most popular express option does not).


goldbergyoni:
@kibertoad Welcome aboard, glad to have you here 👋 

This part indeed should get enriched & modernized not only with Open-API validators but also with class-validator (popular TS decoration based validation).

About the specific package recommendations, did you see this part in our writing-guideliness.md? Didn't check the libs that you mentioned, do they qualify?

image

kibertoad:
I can look into class-based validators and create a separate PR for that.

I've checked with npm, those two libraries are the ones with the most downloads, and qualify on download and update frequency cadence as well.

PR #709, opened 2020-06-05
fix: update spanish translation link to correct url

kevynb:
Change the url link of the spanish translation to point to the correct repo url.


kevynb:
closing in favoir of #710

PR #714, opened 2020-06-07
Docker avoiding secrets - 1st draft

goldbergyoni:
@kevynb @BrunoScheufler @js-kyle Do we want to use the details expander also here? I suggest yes as we will have JS examples that might need to be translated to TS as well


sections/docker/avoid-build-time-secrets.md
```diff
@@ -0,0 +1,92 @@
+# Clean build-time secrets, avoid secrets as args
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+A docker image is not just a bunch of files rather constitutes multiple layers that reveal also what happened during the build time. In a very common scenario, developers need the npm token during build time (mostly for private registries) - this is falsely achieved by passing the token as a build time args. It might seem innocent and safe, however in fact this token can now be fetched from the developers machine Docker history, from the Docker registry and the CI. An attacker who get access to that token, is now capable of writing into the organization private npm registry. There are two more secured alternatives: The flawless one is using Docker --secret feature (experimental as of July 2020) which allows mounting a file during build time only. The second approach is using multi-stage build with args, building and then copying only the necessary files to production. The last technique will not prevent the secrets file from appearing in the local docker history but considered as as secured enough in some organizations.
```
kevynb, line 7: I have an issue with `a bunch of files rather constitutes multiple layers`, but I don't know how to reformulate to make it better...

kevynb: Could we add backticks around --secret in `Docker --secret feature` ? Unless you feel like it's not necessary

kevynb: I think we have an extra `as` here: `considered as as secured`

sections/docker/avoid-build-time-secrets.md
```diff
@@ -0,0 +1,92 @@
...
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+A docker image is not just a bunch of files rather constitutes multiple layers that reveal also what happened during the build time. In a very common scenario, developers need the npm token during build time (mostly for private registries) - this is falsely achieved by passing the token as a build time args. It might seem innocent and safe, however in fact this token can now be fetched from the developers machine Docker history, from the Docker registry and the CI. An attacker who get access to that token, is now capable of writing into the organization private npm registry. There are two more secured alternatives: The flawless one is using Docker --secret feature (experimental as of July 2020) which allows mounting a file during build time only. The second approach is using multi-stage build with args, building and then copying only the necessary files to production. The last technique will not prevent the secrets file from appearing in the local docker history but considered as as secured enough in some organizations.
+
+<br/><br/>
+
+### Code Example – Using Docker mounted secrets (experimental but stable)
```
kevynb, line 11: 👍

sections/docker/avoid-build-time-secrets.md
```diff
@@ -0,0 +1,92 @@
...
+# Deleting the .npmrc within the same copy command will not save it inside the layer, however it can be found in image history
+
+CMD ["node","index.js"]
+```
+
+</details>
+
+<br/><br/>
+
+### Blog Quote: "These secrets aren’t saved in the final Docker"
```
BrunoScheufler, line 84: It sounds like there's something missing here 😄 Not completely sure but I think this talks about the image, so it might be "These secrets aren't saved in the final Docker image", but maybe it's also about the container? 🤔

PR #715, opened 2020-06-08
Avoid process managers - 1st draft

goldbergyoni:


kevynb:
@js-kyle do you mean that we should do a « big » PR when all the docker PRs are merged to fix all the typos we find in the docker section?

sections/docker/restart-and-replicate-processes.md
```diff
@@ -0,0 +1,44 @@
...
+
+<br/><br/>
+
+### One Paragraph Explainer
+
+Docker runtime orchestrators like Kubernetes are really good at making containers health and placement decisions: They will take care to maximize the number of containers, balance them across zones, and can conclude many cluster factors while making these decisions. Goes without words, they identify failing processes (i.e., containers) and restart them in the right place. Despite that, some may tempt to use custom code or tools to replicate the Node process for CPU utilization or restart the process upon failure (e.g., Cluster module, PM2). These local tools don't have the perspective and the data that is available on the cluster level. For example, when the instances resources can host 3 containers and given 2 regions or zones, Kubernetes will take care to spread the containers across zones. This way, in case of a zonal or regional failure, the app will stay alive. On the contrary side, When using local tools for restarting the process, the Docker orchestrator is not aware of the errors and can not make thoughtful decisions like relocating the container to a new instance or zone.
```
kevynb, line 7: What is the meaning of `and can conclude many cluster factors`? I’m not sure conclude is the right word here. Can `deduce`,`determine` or something in the same family fit?

js-kyle:
@kevynb With these big sections we usually prepare a branch with all the bullets completed, rather than one by one as they're ready. So it's probably easier to do spelling/grammar on the final product rather than on each draft? Seems easier, but we can do it however

kevynb:
One way or the other, as long as we try to fix all the mistakes I don’t see any reason to change the way it’s usually done 🙂

sections/docker/restart-and-replicate-processes.md
goldbergyoni: Fixed, thanks

PR #717, opened 2020-06-11
Docker Remove development dependencies - 1st draft

goldbergyoni:


sections/docker/install-for-production.md
```diff
@@ -0,0 +1,54 @@
...
+<details>
+
+<summary><strong>Dockerfile</strong></summary>
+
+```
+
+FROM node:12-slim AS build
+WORKDIR /usr/src/app
+COPY package.json package-lock.json ./
+RUN npm install
```
kevynb, line 41: A quick remark: We could add a comment over that line saying something like: "# Not deleting the cache after npm install"

kevynb: Also, we could have another antipattern example where the `npm clean cache --force` is done in another `RUN` instruction.

goldbergyoni: Added the comment, thanks

PR #719, opened 2020-06-14
Docker memory limit - 1st draft

goldbergyoni:


sections/docker/memory-limit.md
```diff
@@ -0,0 +1,76 @@
...
+    command: ["node index.js --max-old-space-size=400"]
+```
+
+</details>
+
+<br/><br/>
+
+### Kubernetes documentation: "If you do not specify a memory limit"
+
+From [K8S documentation](https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/?fbclid=IwAR2gfRL-3UB2WIbxJNSl3jNdlEJF3rAU5WG049yNorLOVSVngmRY1xhrbqg#if-you-do-not-specify-a-memory-limit)
```
kevynb, line 74: I think we can remote the `fbclid` from the link to avoid tracking: `https://kubernetes.io/docs/tasks/configure-pod-container/assign-memory-resource/#if-you-do-not-specify-a-memory-limit`

sections/docker/memory-limit.md
```diff
@@ -0,0 +1,76 @@
...
...onents to starve. Memory limits also allow the runtime to place a container in the right instance - placing a container that consumes 500MB in an instance with 300MB memory available will lead to failures. Two different options allow configuring this limit: V8 flags (--max-old-space-size) and the Docker runtime. There is confusion around which of those to set, some configure both. The better option would be to set the limit on the Docker command or the Docker runtime (e.g. Kubernetes) as is has a much wider perspective for making the right decisions. First, given this limit, the runtime knows how to scale and create more resources. It can also make a thoughtful decision on when to crash - if a container has a short burst in memory request and the hosting instance is capable of supporting this, Docker will let the container stay alive. Docker also measures the actually used memory and not the allocated part so it can get Killed only when really needed. Last, With Docker the Ops experts can set various production memory configurations that can be taken into account like memory swap. On the contrary, when setting the limit on V8 - each of the described scenarios will not be supported.
```
kevynb, line 7: Great writeup!

- I don't think the part addressing if we can/should not set both flags is crystal clear for readers. Besides `some configure both`, I feel like there is no real explanation. We could insist more on the fact that `--max-old-space-size` should not be set alone but say that one can set both as that can be useful when high memory limits are needed. Is that the case (that one can set both) ? If not, removing the `some configure both` part could help make that clearer or more opinionated (but hey, we are people 😄 ). 

- Given that Kubernetes can run other container engines than docker I think we could use the term `container runtime` instead of docker when referring to k8s.

I suggested phrasing edits but I'm not definitely certain for some of them 🦺 

```suggestion
A memory limit tells the process/container the maximum allowed memory usage - a request or usage beyond this number will kill the process (OOMKill). Applying this is a great practice to ensure one citizen doesn't drink all the juice alone and leaves other components to starve. Memory limits also allow the runtime to place a container in the right instance - placing a container that consumes 500MB in an instance with 300MB memory available will lead to failures. Two different options allow configuring this limit: V8 flags (--max-old-space-size) and the Docker runtime. There is confusion around which of those to set, some configure both. The better option would be to set the limit on the Docker command or the container runtime (e.g. Kubernetes) as it has a much wider perspective for making the right decisions. First, given this limit, the runtime knows how to scale and create more resources. It can also make a thoughtful decision on when to crash - if a container has a short burst in memory request and the hosting instance is capable of supporting this, Docker will let the container stay alive. Docker also measures the actual memory usage and not the allocated part so it can get Killed only when really needed. Last, With Docker the Ops experts can set various production memory configurations that can be taken into account like memory swap. On the contrary, when setting the limit on V8 - each of the described scenarios will not be supported.
```

goldbergyoni: @kevynb Great comments, I'll edit

PR #720, opened 2020-06-14
Docker clean cache - 1st draft

goldbergyoni:


README.md
```diff
@@ -1081,13 +1081,14 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 
 <br/><br/>
 
-## ![✔] 8.1. Clean npm cache
+## ![✔] 8.1. Clean NODE_MODULE cache
 
-**TL;DR:**
+**TL;DR:** After installing dependencies in a container, remove the local cache. It doesn't make any sense to duplicate the dependencies packages to speed future installs since there won't be any further installs - A docker image is created only once. By doing so, using a single line of code, tens of MB, typically 10-50% of the image size are shaved
```
BrunoScheufler, line 1086: Not sure of the `dependencies packages` part is intended, sounds like it should be dependencies _or_ packages, not both 😄

BrunoScheufler, line 1086:
```suggestion
**TL;DR:** After installing dependencies in a container, remove the local cache. It doesn't make any sense to duplicate the dependencies to speed future installs since there won't be any further installs - A Docker image is created only once. By doing so, using a single line of code, tens of MB, typically 10-50% of the image size are shaved off
```

PR #722, opened 2020-06-14
Docker generic advice - 1st draft

goldbergyoni:


sections/docker/generic-tips.md
```diff
@@ -0,0 +1,29 @@
...
+
+**TL;DR:** Updating the local binaries during build (e.g. apt-get update) creates inconsistent images every time it runs and also demands high privilige. Instead use base images that are relatively updated
+
+## ![✔] Tag images using labels
+
+**TL;DR:** Providing meta-data for each image might help Ops professional treat it adequately. For example, include the maintainer name, build date and other information that might prove useful when someone need to reason about an image
+
+## ![✔] Use unprivileged containers
+
+**TL;DR:** Privileged container have the same permissions and capabilities as the root user over the host machine. This is rarely needed and as a rule of thumb should be avoided
```
js-kyle, line 21: Maybe here would could mention how most node images have the `node` user for this purpose

sections/docker/generic-tips.md
```diff
@@ -0,0 +1,29 @@
...
+
+## ![✔] Inspect and verify the final result
+
+**TL;DR:** Sometimes it's easy to overlook side effects in the build process like leaked secrets or unnecessary files. Inspecting the produced image using tools like [Dive](https://github.com/wagoodman/dive) can easily help to identify such issues
```
js-kyle, line 25: It could be nice to add a screenshot/gif of this tool inline

sections/docker/generic-tips.md
```diff
@@ -0,0 +1,29 @@
...
+
+This common Docker guidelines section contains best practices that are standardized among all programming languages and have no special Node.js interpretation
+
+## ![✔] Prefer COPY over ADD command
+
+**TL;DR:** COPY is safer as it copies local files only while ADD supports fancier fetches like downloading binaries from remote sites
+
+## ![✔] Avoid updating the base OS
+
+**TL;DR:** Updating the local binaries during build (e.g. apt-get update) creates inconsistent images every time it runs and also demands high privilige. Instead use base images that are relatively updated
```
BrunoScheufler, line 13:
```suggestion
**TL;DR:** Updating the local binaries during build (e.g. apt-get update) creates inconsistent images every time it runs and also demands elevated privileges. Instead use base images that are updated frequently
```

sections/docker/generic-tips.md
```diff
@@ -0,0 +1,29 @@
...
+
+## ![✔] Tag images using labels
+
+**TL;DR:** Providing meta-data for each image might help Ops professional treat it adequately. For example, include the maintainer name, build date and other information that might prove useful when someone need to reason about an image
```
BrunoScheufler, line 17:
```suggestion
**TL;DR:** Providing metadata for each image might help Ops professionals treat it adequately. For example, include the maintainer name, build date and other information that might prove useful when someone needs to reason about an image
```

kevynb: I think there is a difference between image tag and labels, but it might be a misunderstanding from my part.

For me `tags` are what you seet when you build the image on the cli `-t` or using `docker tag` and that you then see in the registry while `labels` are set in the dockerfile using `LABEL` instructions and are not visible without `inspecting` the image or running different commands (that I do not know since I've never used them).

If that's the case maybe this bullet could explain a bit the difference between the two and provide code examples. I would love to see some just so that I can learn more about the use cases of Labels 👨‍🎓

sections/docker/generic-tips.md
goldbergyoni: @js-kyle Yes!

sections/docker/generic-tips.md
goldbergyoni: Done

sections/docker/generic-tips.md
goldbergyoni: @kevynb Yes, the word 'tag' is confusing here as tags are indeed different than labels and not relevant here. Edited:)

PR #727, opened 2020-07-05
Typo fix: who's -> whose

OrBin:


kevynb:
@allcontributors please add @OrBin for content

PR #731, opened 2020-07-15
2x Docker drafts

js-kyle:
Rough drafts

### DIFF
--- a/sections/docker/image-tags.md
+++ b/sections/docker/image-tags.md
@@ -1,0 +1,18 @@
+# Understand image tags, and use the "latest" tag with caution
+
+### One Paragraph Explainer
+
+
+### Code example:
+
+```
+
+```
+
+
+
+### What Other Bloggers Say
+From the blog [name](https://example.com):
+> A quote
+
+<br/>
--- a/sections/docker/image-tags.md
+++ b/sections/docker/image-tags.md
@@ -1,18 +1,29 @@
-# Understand image tags, and use the "latest" tag with caution
+# Understand image tags vs digests, and use the "latest" tag with caution
 
 ### One Paragraph Explainer
 
+If this is a production situation, and security and stability are important, then just "convenience" is likely not the best deciding factor.
+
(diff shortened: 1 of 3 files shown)


sections/docker/image-tags.md
```diff
@@ -0,0 +1,29 @@
+# Understand image tags vs digests, and use the "latest" tag with caution
+
+### One Paragraph Explainer
+
+If this is a production situation, and security and stability are important, then just "convenience" is likely not the best deciding factor.
```
kevynb, line 5: I might be missing the point, but I think the part about `Understand[ing] image tags vs digests` is missing from the current explanation.

We could write that tagging is understandable by humans but risks being overwritten with a newer image while digests are immutable

sections/docker/image-tags.md
```diff
@@ -0,0 +1,29 @@
...
+$ docker build -t company/image_name:latest .
+# :latest image is updated
+```
+
+### What Other Bloggers Say
+From the blog by [Vladislav Supalov](https://vsupalov.com/docker-latest-tag/):
+> Some people expect that :latest always points to the most-recently-pushed version of an image. That’s not true.
+
+From the [Docker success center](https://success.docker.com/article/images-tagging-vs-digests)
+>
```
kevynb, line 27: Good link! Hard to chose what to quote
Maybe the whole block ?

> ### Tagging is best practice because
>
>    * Human-readable versioning of your images
>    * Easier to run a specific version of an image
>    * Can aid in the process (a release is tagged once it is verified/approved for production)
>
> ### Pinning-by-digest is best practice because
>
>    * Tags are mutable, so there is no guarantee a tag will never change.
>    * It guarantees that every instance of the service is running exactly the same code.
>    * It allows you to put an image through QA/testing and verify that that version of the image is approved to go into production.
>    * You can use Docker Content Trust and sign specific versions of the image.
>    * You can roll back to an earlier version of the image, even if that version was not tagged (or "no longer tagged").
>    * Digests also prevent race-conditions; if a new image is pushed while a deploy is in progress, different nodes may be pulling the images at different times, so some nodes have the new image, some have the old one.
>
> Manually specifying a digest is a bit involved. That's why services automatically resolve tags to digests to make this easy to use.

PR #734, opened 2020-07-24
Leverage caching

kevynb:
I did not talk about `dive`, I need to play with it and use it with a real image


sections/docker/Caching-for-shorter-build-time.md
```diff
@@ -0,0 +1,120 @@
+# Leverage caching to reduce build times
+
+## One paragraph explainer
+
+Docker images are a combination of layers, each instruction in your Dockerfile creates a layer. The docker daemon can reuse those layers between builds if the instructions are identicals or in the case of a `COPY` or `ADD` files used are identical.
+It is important to layout your Dockerfile correctly to reduce the number of moving parts in your build, the less updated instructions should be at the top and the ones constantly changing (like app code) should be at the bottom.
```
goldbergyoni, line 6: Maybe 'The instruction that triggers the longest operation should be at the top to ensure it happens only when really necessary'

sections/docker/Caching-for-shorter-build-time.md
```diff
@@ -0,0 +1,120 @@
+# Leverage caching to reduce build times
+
+## One paragraph explainer
+
+Docker images are a combination of layers, each instruction in your Dockerfile creates a layer. The docker daemon can reuse those layers between builds if the instructions are identicals or in the case of a `COPY` or `ADD` files used are identical.
+It is important to layout your Dockerfile correctly to reduce the number of moving parts in your build, the less updated instructions should be at the top and the ones constantly changing (like app code) should be at the bottom. 
+Rebuilding a whole docker image from cache can be nearly instantaneous if done correctly.
+
+⚠️ If the cache can't be used for a particular layer, all the subsquent layers will be invalidated too. That's why order is important.
```
goldbergyoni, line 9: Consider putting this great sentence somewhere at the opening of the text, it will help to understand what you wrote about the layers

sections/docker/Caching-for-shorter-build-time.md
```diff
@@ -0,0 +1,120 @@
...
+
+#### Have a good .dockerignore file
+
+[**See: On the importance of docker ignore**](/sections/docker/docker-ignore.md)
+
+The docker ignore avoids copying files that could bust our cache logic, like tests results reports, logs or temporary files.
+
+#### Install "system" packages first
+
+If you need to install packages using `apt`,`yum`,`apk` or the likes, this should be one of the first instructions. You don't want to reinstall make,gcc or g++ every time you build your node app.
```
goldbergyoni, line 39: Note that we recommended against updating the OS, maybe include "If you **really** need to install"

sections/docker/Caching-for-shorter-build-time.md
```diff
@@ -0,0 +1,120 @@
...
+It is important to layout your Dockerfile correctly to reduce the number of moving parts in your build, the less updated instructions should be at the top and the ones constantly changing (like app code) should be at the bottom. 
+Rebuilding a whole docker image from cache can be nearly instantaneous if done correctly.
+
+⚠️ If the cache can't be used for a particular layer, all the subsquent layers will be invalidated too. That's why order is important.
+
+![Docker layers](/assests/images/docker_layers_schema.png)
+
+* Image taken from [Digging into Docker layers](https://medium.com/@jessgreb01/digging-into-docker-layers-c22f948ed612) by jessgreb01*
+
+### Rules
```
goldbergyoni, line 15: This deviates a bit from our formatting (having custom section that is no code example or quote) but let's stick to it - it's good. For future bullets, I would stick to the formatting

sections/docker/bootstrap-using-node.md
```diff
@@ -0,0 +1,62 @@
+# Bootstrap container using node command instead of npm
+
+## One paragraph explainer
+
+We are used to see code examples where folks start their app using `CMD 'npm start'`. This is a bad practice. The `npm` binary will not forward signals to your app which prevents graceful shutdown (see #716). If you are using Child-processes they won’t be cleaned up correctly in case of unexpected shutdown, leaving zombie processes on your host. `npm start` also results in having an extra process for no benefit. To start you app use `CMD ['node','server.js']`.
```
goldbergyoni, line 5: Running CMD [node...] will also result in zombies, no? Maybe the advice should be, when no child processes are used you may invoke 'node' directly. Whenever child processes are used, a real process manager should be used like 'TINY' @kevynb

kevynb: It seems to me that if you are not using child-processes, your fine with just running node as PID1 and handling signals. Anyway I added a sentence about it, because that's needed 👍

sections/docker/Caching-for-shorter-build-time.md
kevynb: I partially agree with that, given that sometimes the build code step is the longest one. We don't want to put it at the top in those cases. I'll try to mix and match

PR #745, opened 2020-08-19
Add an example project with dockerfile - 1st draft

kevynb:
I quickly wrote an example project to show usage of the dockerfile. 

I used typescript to have a build step to illustrate multi-stage build. I used koa to change from express ;)

Please let me know what we could add (or contribute to the PR directly)

### DIFF
--- a/sections/examples/dockerfile/Dockerfile
+++ b/sections/examples/dockerfile/Dockerfile
@@ -1,8 +1,17 @@
+# This is a multistage Dockerfile.
+# In the first stage we install system build dependencies, copy project files and build them
+# In the second stage, we start fresh and only copy necessary files. We also purge node_modules devDependencies.
+
+#### Build stage ####
 FROM node:14.8.0-alpine AS build
 
-# Copy dependency information and install all dependencies
+# Install system build dependencies (if needed) at the top ✅ See bullet point #8.8 about caching
+RUN apk add --update --no-cache bash make gcc g++ lcms2-dev libpng-dev autoconf automake
+
+# Only copy node dependency information and install all dependencies first
 COPY --chown=node:node package.json package-lock.json ./
 
+# Install packages using the lockfiles as source of truth ✅ See bullet point #8.5 about npm ci
 RUN npm ci
 
 # Copy source code (and all other relevant files)
@@ -11,8 +20,9 @@ COPY --chown=node:node src ./src
 # Build code
 RUN npm run build
 
-# Run-time stage
-FROM node:14.8.0-alpine
+#### Run-time stage ####
+# ✅ See bullet point #8.10 about smaller docker base images
+FROM node:14.8.0-alpine as app
 
 # Set non-root user and expose port 3000
 USER node
@@ -20,12 +30,13 @@ EXPOSE 3000
 
 WORKDIR /home/node/app
 
-# Copy dependency information and build results from previous stage
+# Copy dependency information and build output from previous stage
 COPY --chown=node:node --from=build package.json package-lock.json ./
 COPY --chown=node:node --from=build node_modules ./node_modules
 COPY --chown=node:node --from=build dist ./dist
 
-# Clean dev packages
-RUN npm prune --production
+# Clean dev dependencies ✅ See bullet point #8.5
+RUN npm prune --production && npm cache clean --force
 
+# ✅ See bullet point #8.2 about avoiding npm start
 CMD [ "node", "dist/app.js" ]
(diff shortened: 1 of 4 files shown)


sections/examples/dockerfile/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+COPY --chown=node:node package.json package-lock.json ./
+
+RUN npm ci
+
+# Copy source code (and all other relevant files)
+COPY --chown=node:node src ./src
+
+# Build code
+RUN npm run build
+
```
osher, line 13: @goldbergyoni  - I love your work. 

I would add here `RUN npm run test`, goes with `"test": "node test.js"` in `package.json` and a `test/index.js` file that launches the server and hit it on `'/'`. 
You can also do it with `mocha` or `jest`, and use a `beforeAll` hook to launch the server and `afterAll` hook to terminate it. Checkout [e2e-helper](https://github.com/osher/e2e-helper)

A lot of value for not that much complexity :)

rockrotem (commented):
I would add build tools example step:
RUN apk add --update curl && \
    apk add --no-cache --virtual .build-deps bash make gcc g++ lcms2-dev libpng-dev autoconf automake && \
npm run build && \
apk del .build-deps

This will install build tools and delete them after to keep docker slim :-)

sections/examples/dockerfile/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Run-time stage
+FROM node:14.8.0-alpine
```
goldbergyoni, line 15: Show installation of private packages by using --secret mount

sections/examples/dockerfile/Dockerfile
goldbergyoni: @osher Thanks, happy to hear that words!

goldbergyoni (approved):

sections/examples/dockerfile/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Copy dependency information and build results from previous stage
+COPY --chown=node:node --from=build package.json package-lock.json ./
+COPY --chown=node:node --from=build node_modules ./node_modules
+COPY --chown=node:node --from=build dist ./dist
+
+# Clean dev packages
+RUN npm prune --production
+
+CMD [ "node", "dist/app.js" ]
```
goldbergyoni, line 31: @kevynb To make it even more attractive, you may put nearby each relevant line the reference to the BP. Like this:

// ✅ BP 8.6

sections/examples/dockerfile/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "node-app-with-docker",
+  "version": "1.0.0",
+  "description": "An example node app that uses docker to be built",
+  "main": "src/app.ts",
+  "scripts": {
+    "run": "dist/app.js",
```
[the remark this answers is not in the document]
kevynb: Totally! It was a mistake

sections/examples/dockerfile/Dockerfile
kevynb: I started doing it but it wasn't easy to understand what was happening without the `docker build` command alongside. So I skipped it for now. Let's do it when we create a full example app and add a ci step

sections/examples/dockerfile/Dockerfile
kevynb: Nice! I did that

sections/examples/dockerfile/Dockerfile
kevynb: @osher we had the debate on a previous issue about that. 

It is possible to run the tests in the Dockerfile but it's not what it is meant for. It is better practice to build an image with only the first stage (using the flag `--target build`) and then run testing, linting and other "checks" on this image. 
This allows to parallelize runs (each tool can run in its own container) and is more inline with the vision that Dockerfiles are for building images and docker containers are for running commands.
Depending on which tool you are using for your builds, it's also easier to see which step of the build failed since you have the step name instead of having to look at the Dockerfile build output to understand.

However we'll keep your example in mind when we'll work on a complete app example 👍

kevynb:
Thanks for the suggestion @rockrotem I added a line to install system dependencies. However, as explained in [the bullet point about caching ](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/docker/use-cache-for-shorter-build-time.md) it's better to install system packages at the top of the file. 
If we install apk packages and node modules in the same layer, every time we will change our `package.json` we will have to reinstall all the apks even though they did not change.

Also, since we are in a multi-stage build, there is no real need to cleanup our mess, these stage will be discarded eventually.

goldbergyoni (approved):

PR #758, opened 2020-08-25
docs: add new section: 2.12 Always await promises before returning

Alexsey:
TL;DR: In order to have full stacktraces
1. Always prefer `return await` over `return` when returning promises
2. Any function that is returning a promise must be declared as `async`

[Issue](https://github.com/goldbergyoni/nodebestpractices/issues/737)


ericblade:
... isn't return await useless?

Alexsey:
@ericblade in case of v8 they are essential in reducing stress level :)

you can try yourself the code from the issue and see for yourself

ericblade:
"return await" however, is a bit of a problem, and would be better showcased using "const x = await whatever(); return x;"  ... for when you need to deal with a stack trace problem, I guess.  I don't see where there's really much of a problem, though, it still ends up going back to the right place eventually.  

https://jakearchibald.com/2017/await-vs-return-vs-return-await/

also

https://eslint.org/docs/rules/no-return-await

Alexsey:
@ericblade this proposal is about how to return promises from functions. If you split `return await` with an intermediate variable you will just avoid the situation of returning a promise from an async function and the rule would sound like "never return promises from functions". Yes, this is also a way to have a good async stacktrace but the price you would need to pay for the same result is higher: for example, you would need to forbid implicit return in arrow functions if the function returns a promise

the article you have mentioned is very good and absolutely correct as for 2017 - I've sent it to a lot of people when it appeared :) But in 2017 there were not async stacktraces at all. Then almost two years ago v8 team had come with there "zero-cost async stacktrace" implementation that made a difference between `return` and `return await`

As for eslint rule, first of all, it has also appeared before "zero-cost async stacktrace" in v8. In second, this thing is specific to v8 and eslint is not about v8 - it's about EcmaScript in general while Node Best Practices are only about v8. In third, in the rule, there is a line in "When Not To Use It": "If you want the functions to show up in stack traces". You can read the discussion about it here https://github.com/eslint/eslint/issues/11878 - authors of eslint were actually good with adding more info to the rule description but the topic starter decided to drop the idea for his own reasons

goldbergyoni:
@Alexsey Thanks again, I'm OOO until mid next week and I'll be sure to review then

sections/errorhandling/returningpromises.md
```diff
@@ -0,0 +1,94 @@
...
+Error: boom!
+    at asyncThrow ([...])
+    at async asyncFn ([...])
+```
+
+where the middle of the stacktrace is missing `syncFn`
+
+Notes:
+1. One another reason why async stacktrace has such tricky implementation is the limitation that stacktrace
+must always be build synchronously, on the same tick of event loop <span id="a1">[¹](#1)</span>
```
kevynb, line 81: I think it should be `always be built synchronously` here

README.md
```diff
@@ -228,6 +228,20 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 
 <p align="right"><a href="#table-of-contents">⬆ Return to top</a></p>
 
+## ![✔] 2.12 Always await promises before returning
+
+**TL;DR:** Always do `return await` when returning a promise. If a function returns a promise, that  
+function must be declared as async function and explicitly await the promise before returning it
+
+**Otherwise:** The function would be missed in stacktraces and errors would look like they came out of
```
kevynb, line 236: I think it should be `would be missing in stacktraces`

README.md
```diff
@@ -228,6 +228,20 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 
 <p align="right"><a href="#table-of-contents">⬆ Return to top</a></p>
 
+## ![✔] 2.12 Always await promises before returning
+
+**TL;DR:** Always do `return await` when returning a promise. If a function returns a promise, that  
+function must be declared as async function and explicitly await the promise before returning it
+
+**Otherwise:** The function would be missed in stacktraces and errors would look like they came out of  
+something like `process.nextTick` in to a random place in code
```
kevynb, line 237: What about rephrasing with something like `from a random place in node internal code` ?

sections/errorhandling/returningpromises.md
```diff
@@ -0,0 +1,94 @@
+# Returning promises
+
+<br/>
+
+### One Paragraph Explainer
+
+The mechanisms behind sync function stacktrace and async function stacktrace in v8 implementation are quite different:
+sync stacktrace is based on **stack** provided by operation system Node.js is running on (just like in most programming
```
kevynb, line 8: We could use the plural when talking about the stacktraces, something like: `sync stacktraces are based on the stack provided by`

I think we should say `the operating system` instead of `operation` here and in the other sentences

sections/errorhandling/returningpromises.md
```diff
@@ -0,0 +1,94 @@
...
+languages). When async function is executing, the **stack** of operation system is popping it out as soon as the
```
kevynb, line 9: Here I think we could write either `When an async function is executing` or `When async function are executing` (but this one needs little changes in the rest of the text).

README.md
```diff
@@ -228,6 +228,20 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 
 <p align="right"><a href="#table-of-contents">⬆ Return to top</a></p>
 
+## ![✔] 2.12 Always await promises before returning
+
+**TL;DR:** Always do `return await` when returning a promise. If a function returns a promise, that  
+function must be declared as async function and explicitly await the promise before returning it
```
kevynb, line 234: We might want to add backquotes around `await` and `async` here too

kevynb (commented):
Thanks for taking the time to write this BP @Alexsey, great writeup! 

Only suggested a few minor edits

sections/errorhandling/returningpromises.md
Alexsey: I have tried to pluralize "stacktraces", but with it I needed to pluralize "functions", "stacks", "promises", "chains" etc., and when everything is plural it's not clear that all relations between entities are always 1-to-1, so the text becomes less univocal and I decide to pluralize only the first part of the first sentence

README.md
Alexsey: I don't want to have two "random" one after another, and, actually nothing there is random :) so I have rephrased it a little bit differently. Hope new wording sounds better

goldbergyoni:
@Alexsey Just a note that I plan to chime-in very soon, just a busy period

Alexsey:
There is one edge case that I think worth to discuss: short functions high order functions
```js
const users = await Promise.all(usersIds.map(id => getUser(id, options))
```
and non-anonimous high order functions
```js
const users = await Promise.all(usersIds.map(getUser))
```
In case if `getUser` would throw the `map` would not be in a stacktrace. Considering the fact that `Promise.all` would be in the stacktrace, does it really worth banning this kind of code?

README.md
goldbergyoni: This otherwise sentence is technically true but not simple enough. Maybe:

'The function would not be part of the StackTrace and it will be harder to follow an error path without some of the calls chain'

goldbergyoni:
@Alexsey 

Thanks again for bringing this hop topic 🔥

I read all the related materials, couple of thoughts:

1. No clear cut here - I'm not sure that this is a case for an explicit recommendation as one might favor performance over observability (see bullet 3' 👇), and going against an ESLint rule (partially) is something to consider. Unless, we can clarify that these guides are completely wrong (see bullet 3' 👇). In any case, almost all the text and great examples that you brought are still relevant, we would just need to specify the trade-off. Probably tweak and add few sentences

2. Formatting - You have great English writing skills for non-native speaker, still the texts and formatting should be revised. By the way, I'm also not a native English speaker, like you, I struggle also in my writing here, @BrunoScheufler @js-kyle helps me a lot in sharpening my writing (+200 PR that fixed TYPOs and grammar issues 😅). I'll share more detailed thoughts on this soon, once we conclude our main advice here

3. Tech question - The ESLint rule states that:

> Keeps the current function in the call stack until the Promise that is being awaited has resolved, at the cost of an extra microtask before resolving the outer Promise

You wrote that:

> had come with there "zero-cost async stacktrace" implementation that made a difference between return and return await

Isn't the modern asyn stacktrace keep the function in the call stack until the outer promise is resolved?

Alexsey:
@goldbergyoni thank you for the review!

You are correct - we need to add some notes about performance tradeoff and I already know what to write

As for your question, I don't see any conflict between ESLint description any what I've written: after "zero-cost async stacktrace" has been introduced `return await`would keep the current function in the call stack until the Promise that is being awaited would be resolved while `return` would *not* keep the current function in the call stack. Not all modern sync stacktraces works this way: Firefox doesn't require `return await` for having good async stacktraces. Please, tell me if you still see some collisions

Alexsey:
Added `Tradeoff` section to `returningpromises.md`, changed `Otherwise` sentence as suggested, and improve Russian translation

The PR is ready for the review!

goldbergyoni (approved):
[I mistakenly approved instead of 'Request changes', please treat this accordingly]

@Alexsey Great progress, I believe we can merge this in days and it will be valuable even to experienced developers

I left a few comments, here is a summary:

1. Mandatory: The 'one paragraph' should start with a simpler explanation. The advanced stuff are great but should come later. I proposed a skeleton for the simple text in my comment

2. Mandatory: Revise the English grammar, this should be done after the text is done and can be relatively short 

@BrunoScheufler @js-kyle Would you be able to help with this?

3. Optional: Other comments which I left inside, take whatever makes sense to you

goldbergyoni:
@Alexsey I'm just recaping here the current status, LMK if we/I can help somehow with pushing this:

I mistakenly approved instead of 'Request changes', requested edits:

1. Mandatory: The 'one paragraph' should start with a simpler explanation. The advanced stuff are great but should come later. I proposed a skeleton for the simple text in my comment

2. Mandatory: Revise the English grammar, this should be done after the text is done and can be relatively short

3. Optional: Other comments which I left inside, take whatever makes sense to you

Alexsey:
@goldbergyoni I'm going to go over all of your comments and also add some small things I have think about, but right now I'm our of time :( I expect to get back to this next weekend. The PR is not forbidden - I'm having some thoughts about it almost every day :)

goldbergyoni:
> @goldbergyoni I'm going to go over all of your comments and also add some small things I have think about, but right now I'm our of time :( I expect to get back to this next weekend. The PR is not forbidden - I'm having some thoughts about it almost every day :)

@alexsey This is the OSS land, there are no clocks, no time, no gravity, no apologies, only good will 🙂. In other words, take as much time as needed

Alexsey:
The PR is ready for the next iteration of review: I went over all comments, added a simplified one paragraph explanation, split all code examples into "Anti-pattern" + "the right way to go" and added one new anti-pattern when usage of async callback is leading to missing frames in stackteaces

goldbergyoni (changes requested):
@Alexsey Thanks for making this even better and for understanding our great care for the content quality. The home page text is great and the advanced explanations are really outstanding.

**We're 10 min of work from approving this PR, 4 fundamental asks:**
1. One paragraph should start with a simple explanation - There should be a simpler and actionable explanation for **newbie average developers**. See an example of how simple it should be below

2. History is out of scope - The one paragraph contains history of changes in Node, our future reader in 2022 won't care about this, it neither exists in other bullets. You may exclude this or move to the advanced explanation

3. Advanced should be at the bottom - Our format is always 'One paragraph' and then code examples. There are no other bullets with advanced explanations like your amazing drill down. To make this consistent, at least let's place the advanced text at the bottom including everything that is not 'one paragraph' or 'code examples'

4. Code examples should follow our format - See attached an image that exemplify the formatting of code examples, they all start with the same text. It's not mandatory to put the JS/TS expander, but the title should be consistent

image


**Dead-simple explanation for the opening of the one paragraph:
'"When function A calls async function B, make A also async and await on B's response, otherwise A won't appear in the stacktrace in case of error. This might confuse the developer who tries to understand the flow..."

goldbergyoni:
Hey @Alexsey , Just checking whether you need anything from me or good to make the slight adjustments. This is by no mean aim to rush you, take your time

Alexsey:
Thank you for reminding - you have guessed a good moment :) I finally have some spare time to finish with this one - 90% will make the PR during the next 3 days, hope even today's evening.

Alexsey:
Simplified "One paragraph", add collapsable blocks to code snippets, moved history to a separate paragraph called "Why return await was considered as anti-pattern in the past" and move it with "Advanced explanation" below code examples

I didn't remove history because there is a lot of people who consider `return await` as anti-pattern, and I think it's important to explain that this information is not wrong but obsolete

goldbergyoni (approved):
LGTM

goldbergyoni:
@Alexsey That's an epic PR, great idea and execution 🏆

Let's promote this:
1. I'm gonna tweet about. Do you happen to have a  Twitter handle?
2. Put it in our news section
3. Add you to the contributors list - @all-contributors please add @Alexsey for content

goldbergyoni:
@all-contributors please add @Alexsey for content

Alexsey:
@goldbergyoni Cool, really glad we have finally done this, and thank you for your intention to promote!

Unfortunately, I don't have a Twitter account :(

goldbergyoni:
@Alexsey That's fine, I'll tweet anyway and promote this great content!

goldbergyoni:
@Alexsey Would you like to chime-in here?

https://www.reddit.com/r/node/comments/k1x8p1/a_new_best_practice_212_await_before_returning/

Alexsey:
@goldbergyoni now I will probably even register on Reddit =) Thank you for the link!

PR #759, opened 2020-08-26
Use npm ci command in Docker builds

bttger:
Also added the hint in chapter 5.19 to use the `npm ci` command in automated environments.

Maybe this should also be updated in the *in detail section* [Multi Stage Builds with Docker](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/docker/multi_stage_builds.md), but this example uses yarn, which would be `yarn install --frozen-lockfile` then. Ideally with a short explanation that both commands can be fine, depending which package manager you use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1085,7 +1085,7 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 
 ## ![✔] 8.1 Use multi-stage builds for leaner and more secure Docker images
 
-**TL;DR:** Use multi-stage build to copy only necessary production artifacts. A lot of build-time dependencies and files are not needed for running your application. With multi-stage builds these resources can be used during build while the runtime environment contains only what's necessary. Multi-stage builds are an easy way to get rid of overweight and security threats
+**TL;DR:** Use multi-stage build to copy only necessary production artifacts. A lot of build-time dependencies and files are not needed for running your application. With multi-stage builds these resources can be used during build while the runtime environment contains only what's necessary. Multi-stage builds are an easy way to get rid of overweight and security threats. Since Docker is often used in continous integration environments it is recommended to use the `npm ci` command (instead of `npm install`). It is faster, stricter and reduces inconsistencies by using only the versions specified in the package-lock.json file. See [here](https://docs.npmjs.com/cli/ci.html#description) for more info.
 
 **Otherwise:** Larger images will take longer to build and ship, build-only tools might contain vulnerabilities and secrets only meant for the build phase might be leaked.
 
@@ -1095,7 +1095,7 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 FROM node:14.4.0 AS build
 
 COPY . .
-RUN npm install && npm run build
+RUN npm ci && npm run build
 
 FROM node:slim-14.4.0
 
@@ -1103,7 +1103,7 @@ USER node
 EXPOSE 8080
 
 COPY --from=build /home/node/app/dist /home/node/app/package.json /home/node/app/package-lock.json ./
-RUN npm install --production
+RUN npm ci --production
 
 CMD [ "node", "dist/app.js" ]
 ```
--- a/README.md
+++ b/README.md
@@ -1085,7 +1085,7 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 
 ## ![✔] 8.1 Use multi-stage builds for leaner and more secure Docker images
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1085,7 +1085,7 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 
 ## ![✔] 8.1 Use multi-stage builds for leaner and more secure Docker images
 
-**TL;DR:** Use multi-stage build to copy only necessary production artifacts. A lot of build-time dependencies and files are not needed for running your application. With multi-stage builds these resources can be used during build while the runtime environment contains only what's necessary. Multi-stage builds are an easy way to get rid of overweight and security threats
+**TL;DR:** Use multi-stage build to copy only necessary production artifacts. A lot of build-time dependencies and files are not needed for running your application. With multi-stage builds these resources can be used during build while the runtime environment contains only what's necessary. Multi-stage builds are an easy way to get rid of overweight and security threats. Since Docker is often used in continuous integration environments it is recommended to use the `npm ci` command (instead of `npm install`). It is faster, stricter and reduces inconsistencies by using only the versions specified in the package-lock.json file. See [here](https://docs.npmjs.com/cli/ci.html#description) for more info.
```
kevynb, line 1088: I would add this line to `docker/multi_stage_builds.md` instead of having it here. The bullet point here is long enough and there's already BP 5.19 that you already enriched nicely :)

README.md
```diff
@@ -734,9 +734,9 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 5.19. Install your packages with `npm ci`
+## ![✔] 5.19. Install your packages with `npm ci` 
 
-**TL;DR:** You have to be sure that production code uses the exact version of the packages you have tested it with. Run `npm ci` to do a clean install of your dependencies matching package.json and package-lock.json.
+**TL;DR:** You have to be sure that production code uses the exact version of the packages you have tested it with. Run `npm ci` to do a clean install of your dependencies matching package.json and package-lock.json. Using this command is recommended in automated environments such as continuous integration pipelines. It is stricter by throwing an error if the versions in package.json and package-lock.json do not match. Since it can be significantly faster than `npm install`, it can also save you CI pipeline minutes. See [here](https://docs.npmjs.com/cli/ci.html#description) for more info.
```
kevynb, line 739: Cool! Can you also add these pertinent info to the `docker/installpackageswithnpmci.md` page?

kevynb (commented):
Hey mate, thanks for contributing! That's good insights you add here 👍

README.md
bttger: I agree, have put it to the `docker/multi_stage_builds.md` file and updated the code examples

README.md
bttger: I haven't found this file. Do you mean [`docker/install-for-production.md`](https://github.com/bttger/nodebestpractices/blob/master/sections/docker/install-for-production.md)? This section already mentions the `npm ci`command and also quotes the npm docs at the bottom.

goldbergyoni, line 739: @bttger The content is great, just make the TLDR shorter, choose the core messages, keep the rest to the 'more info' page

README.md
```diff
@@ -1095,15 +1095,15 @@ Bear in mind that with the introduction of the new V8 engine alongside the new E
 FROM node:14.4.0 AS build
 
 COPY . .
-RUN npm install && npm run build
+RUN npm ci && npm run build
```
goldbergyoni, line 1098: @BrunoScheufler You wrote this so just cc'ing you, I guess 'ci' is better here and more consistent with the rest of our recommendations, right?

goldbergyoni (changes requested):
Great content! Left minor change request

goldbergyoni:
@bttger Just saying welcome ☀️. Left minor review suggestion.

README.md
bttger: Ah I found it, it was under the `production` directory. Going to shorten the TL;DR and put it into the "more info" page.

bttger: Thanks, I deleted following part

> It is stricter by throwing an error if the versions in package.json and package-lock.json do not match. Since it can be significantly faster than `npm install`, it can also save you CI pipeline minutes. See [here](https://docs.npmjs.com/cli/ci.html#description) for more info.

since these points are already stated in the "more info" page. I think the hint that the `ci` command should be used in automated environments should stay in the TL;DR.

goldbergyoni (approved):
LGTM:)

README.md
BrunoScheufler: Sure, we can change it to npm ci 👍

goldbergyoni:
@bttger I thank you again for this important enhancements. It's all here about micro-optimizations, all of them together makes the difference between mediocre software and a good one:)

@all-contributors please add @bttger  for content

PR #760, opened 2020-08-27
All privates packages options

aditrades:


kevynb:
Hello @aditrades  and thanks for taking the time to contribute to our project 👏 

I don't really see the purpose of this PR. What is the goal of adding all of the `npm install` options? 
I'm not sure that it is relevant to this bullet point 😕 You might be able to explain though 🙂

aditrades:
> Hello @aditrades and thanks for taking the time to contribute to our project 👏
> 
> I don't really see the purpose of this PR. What is the goal of adding all of the `npm install` options?
> I'm not sure that it is relevant to this bullet point 😕 You might be able to explain though 🙂

Just wanted to mention that there are more than three options to use privates packages.
The npm install variations summaries all the possibilities.
For example: for private packages, we are using git repo and tag and etc.
Just wanted to mention that private modules, private registry and local npm packages are not the only options for private packages.
(feel free to edit if you think my approach is not fully understood)

kevynb:
I see, while it’s nice as docs it doesn’t fit the purpose of best practices thus I'm going to close this PR.

However if you feel like updating our BP `5.4. Lock dependencies` instead we would be grateful 🙂 

It needs a refresh, in there we're still talking about `npm shrinkwrap` but now it's better to use the `package-lock.json` file instead.
See also `sections/production/lockdependencies.md` if you're interested to help for that.

PR #791, opened 2020-11-05
Add Winston logger to 2.7 (loggers)

denisecase:
Adding Winston as a top option under 2.7 loggers.  Reference https://www.npmtrends.com/winston-vs-pino-vs-bole-vs-log4js

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ Read in a different language: [![CN](/assets/flags/CN.png)**CN**](/README.chines
 
 ## ![✔] 2.7 Use a mature logger to increase error visibility
 
-**TL;DR:** A set of mature logging tools like [Pino](https://github.com/pinojs/pino) or [Log4js](https://www.npmjs.com/package/log4js), will speed-up error discovery and understanding. So forget about console.log
+**TL;DR:** A set of mature logging tools like [Winston](https://github.com/winstonjs/winston), [Pino](https://github.com/pinojs/pino) or [Log4js](https://www.npmjs.com/package/log4js), will speed-up error discovery and understanding. So forget about console.log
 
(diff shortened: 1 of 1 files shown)


goldbergyoni:
@denisecase Winston should be in the list. I would suggest that it should appear under Pino because:

1. We conducted a benchmark last year which showed that Pino was tremendously faster
2. Winston main examples show logging to file system which encourages a bad practice (logging not to stdout)

I'm curious to hear your thoughts

@kevynb Thoughts on this?

josh-hemphill:
^
I really like [pino's assertion in their readme](https://www.npmjs.com/package/pino#transports--log-processing). 
Though I noticed that the detailed explanation instead puts Winston first; should the detailed explanation also be updated?
Perhaps the Readme should only assert that you need to adopt a single structure for all your logging, so that (as pino points out) an external system can predictably parse and consume your logging.

denisecase:
Adding Winston after Pino sounds better - thanks for the info!

kevynb:
@goldbergyoni I second you on that. 

@denisecase I dug out [the issue where spoke about that](https://github.com/goldbergyoni/nodebestpractices/issues/684) for you in case you want more context.

@denisecase or @josh-hemphill Would you care to make a PR to change the to the detailed explanation?
If not, I'll find some time to do it later.

denisecase:
@kevynb The discussion is excellent and I agree Winston is not best practices. That's why this guide is so helpful.  I'd recommend we remove all Winston example code and change the opening to:  

We all love console.log but obviously, a reputable and persistent logger like Pino (a [newer option focused on performance](https://github.com/goldbergyoni/nodebestpractices/issues/684)) is mandatory for serious projects.

denisecase:
This is no longer best practice - RP can be closed.

PR #792, opened 2020-11-13
[Chinese]add avoid-build-time-secrets.chinese.md

MattJin:
add avoid-build-time-secrets.chinese.md


BrunoScheufler (approved):
🙌 thank you for contributing once again 😄

PR #835, opened 2021-01-02
Revise blacklist/whitelist terminology

skinofstars:
Revise questionable terminology, i.e. blacklist/whitelist

https://tools.ietf.org/id/draft-knodel-terminology-00.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,15 +883,15 @@ All statements above will return false if used with `===`
 
 <br/><br/>
 
-## ![✔] 6.11. Support blacklisting JWTs
+## ![✔] 6.11. Support blocklisting JWTs
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A2-Broken_Authentication" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A9:Broken%20Authentication%20-green.svg" alt=""/></a>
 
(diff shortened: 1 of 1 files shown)


elektronik2k5 (approved):

README.md
```diff
@@ -1054,7 +1054,7 @@ All statements above will return false if used with `===`
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A6-Security_Misconfiguration" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A6:Security%20Misconfiguration%20-green.svg" alt=""/></a>
 
-**TL;DR:** Precautions should be taken to avoid the risk of accidentally publishing secrets to public npm registries. An `.npmignore` file can be used to blacklist specific files or folders, or the `files` array in `package.json` can act as a whitelist.
+**TL;DR:** Precautions should be taken to avoid the risk of accidentally publishing secrets to public npm registries. An `.npmignore` file can be used to ignore specific files or folders, or the `files` array in `package.json` can act as an allow list.
```
kevynb, line 1057: I think I would replace `an allow list` with something different. How does `an include-only list` sounds ? I'm open to suggestions, I just find `allow list` weird for that exact use case.

kevynb:
Also, there's a newer version  of this draft: https://tools.ietf.org/id/draft-knodel-terminology-04.html

README.md
goldbergyoni: @kevynb Isn't the PR text follow the recommended and standard wording? I took a look at the newer version and it seems that allow/block are still the recommended new standard

kevynb:
Ah, sorry my point was not clear at all! I just wanted to have a link to the latest version of the draft, not point out a difference in terminology.

README.md
kevynb: Alright, maybe it's because it's a new terminology and I'm not used to it :) I find it weird that's all, I'm not opposed to merging because of that. I should go and comment on the ietf draft instead 🙃

kevynb (approved):

PR #839, opened 2021-01-07
Define DAL in its first use

nosvalds:


kevynb (approved):

kevynb:
@all-contributors please add @nosvalds for content

PR #868, opened 2021-02-03
fix the bad code example in same line curly braces

rluvaton:


elektronik2k5 (approved):
@rluvaton thanks for fixing!
@goldbergyoni how about we recommend using Prettier instead of all these opinionated rules?

rluvaton:
> @goldbergyoni how about we recommend using Prettier instead of all these opinionated rules?

@elektronik2k5 if we just write to use prettier the reason such specific rules are important will disappear

And as of Prettier documentation, Prettier **is**

> An opinionated code formatter

elektronik2k5:
@rluvaton you're right that both are opinionated, of course. However, Prettier is pretty much considered the industry standard (at least in OSS) and is a very good default. And we already recommend it here anyways. 
So perhaps the the best thing to do is to mention that if you use Prettier (as suggested) then there's no need to format by hand.

rluvaton:
> So perhaps the the best thing to do is to mention that if you use Prettier (as suggested) then there's no need to format by hand.

I think it's a good idea

goldbergyoni:
@rluvaton @elektronik2k5 Agree with both - It's important to understand WHY these rules are important, and practically one should enforce this using Prettier

Where should we add that prettier recommendation?

goldbergyoni:
@rluvaton @elektronik2k5 Agree with both - It's important to understand WHY these rules are important, and practically one should enforce this using Prettier

goldbergyoni (approved):
LGTM

goldbergyoni:
@all-contributors please add @rluvaton for content

goldbergyoni:
@all-contributors please add @elektronik2k5  for content

rluvaton:
> Where should we add that prettier recommendation?

Right after 3.1 ESLint,
Should we also specify to use prettier plugin with ESLint?

PR #884, opened 2021-02-11
Update star count

bl-ue:
60,517 currently, to be exact, the the latest star by me.


bl-ue:
Maybe it's not so good an idea to do this so soon, ATM 90 more stars were added for a total of 60,607 :smile:

PR #898, opened 2021-02-24
Update test example to follow the AAA rule

rluvaton:
The test examples have some anti patterns 

> Anti Pattern: no separation, one bulk, harder to interpret
> 
> From [AAA](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/testingandquality/aaa.md#code-example--anti-pattern-no-separation-one-bulk-harder-to-interpret)

### DIFF
--- a/sections/testingandquality/avoid-global-test-fixture.md
+++ b/sections/testingandquality/avoid-global-test-fixture.md
@@ -11,13 +11,15 @@
 ### Code example: each test acts on its own set of data
 ```javascript
 it('When updating site name, get successful confirmation', async () => {
-  //test is adding a fresh new records and acting on the records only
+  //Arrange - test is adding a fresh new records and acting on the records only
   const siteUnderTest = await SiteService.addSite({
     name: 'siteForUpdateTest'
   });
 
+  //Act
   const updateNameResult = await SiteService.changeName(siteUnderTest, 'newName');
 
+  //Assert
   expect(updateNameResult).to.be(true);
 });
 ```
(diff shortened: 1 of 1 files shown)


sections/testingandquality/avoid-global-test-fixture.md
```diff
@@ -15,7 +15,9 @@ it('When updating site name, get successful confirmation', async () => {
   const siteUnderTest = await SiteService.addSite({
     name: 'siteForUpdateTest'
   });
+
```
goldbergyoni, line 18: @rluvaton Why not AAA it appropriately? Shall we create AAA eslint rule?

AAA example:

image

goldbergyoni (changes requested):
@rluvaton Great improvement 🔥

p.s. Shall we create AAA eslint rule?

sections/testingandquality/avoid-global-test-fixture.md
rluvaton: Is my last commit resolve what you requested?

rluvaton:
> @rluvaton Great improvement 🔥

Thank you, happy to contribute

> p.s. Shall we create AAA eslint rule?

I think it's a good idea, how would you see the rule?

Will it warn when no comment Arrange

goldbergyoni (approved):
LGTM 🦄

sections/testingandquality/avoid-global-test-fixture.md
goldbergyoni: Yes! 🔥

goldbergyoni:
@rluvaton 

> How do you see this rule in action?

Raw thoughts - There must be Arrange/Act/Assert in the test + at least one line with code or space between them. Will it warn or show error - This is always configurable in eslint. 

Here is an example of eslint rule related with testing:

https://github.com/lo1tuma/eslint-plugin-mocha/blob/master/lib/rules/valid-test-description.js

rluvaton:
> There must be Arrange/Act/Assert in the test

I think it won't complain when the user wrote _Given/When/Then_ as they are practically the same (although this [SO answer](https://softwareengineering.stackexchange.com/a/308540/335772) says differently)

> at least one line with code or space between them.

What do you mean by *space between them.*?

> Will it warn or show error - This is always configurable in eslint.

Isn't there a default option?

PR #943, opened 2021-03-31
correct an example code in 3.3

Two-Jay:
i've corrected the example code in part 3.3 on the Korean translated version,  according to the below StackOverFlow post.

### DIFF
--- a/README.korean.md
+++ b/README.korean.md
@@ -259,7 +259,8 @@ function someFunction() {
 }
 
 // 나쁜 예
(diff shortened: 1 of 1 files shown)


goldbergyoni:
Yo @Two-Jay Welcome 🌻

Which SO post? Can you clarify plz - It this content exists in every translation?

goldbergyoni:
@Two-Jay Ping:)

PR #946, opened 2021-04-02
docs: convert var to const

rluvaton:


elektronik2k5 (approved):

goldbergyoni:
Another great catch by @rluvaton

PR #947, opened 2021-04-02
Update 6.8 and add `argon2`

maestroevn:
See OWASP cheat sheet - https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html

### DIFF
--- a/README.md
+++ b/README.md
@@ -844,11 +844,11 @@ Also known as correlation id / transit id / tracing id / request id / request co
 
 <br/><br/>
 
-## ![✔] 6.8. Protect Users' Passwords/Secrets using bcrypt or scrypt
+## ![✔] 6.8. Protect Users' Passwords/Secrets using argon2id, bcrypt or scrypt
 
 <a href="https://www.owasp.org/index.php/Top_10-2017_A2-Broken_Authentication" target="_blank"><img src="https://img.shields.io/badge/%E2%9C%94%20OWASP%20Threats%20-%20A9:Broken%20Authentication%20-green.svg" alt=""/></a>
 
-**TL;DR:** Passwords or secrets (e.g. API keys) should be stored using a secure hash + salt function like `bcrypt`,`scrypt`, or worst case `pbkdf2`.
+**TL;DR:** Passwords or secrets (e.g. API keys) should be stored using a secure hash + salt function like `argon2id`, `bcrypt`,`scrypt`, or worst case `pbkdf2`. The winner of password hashing competition is argon2 algorithm and recommended method for password storage by OWASP.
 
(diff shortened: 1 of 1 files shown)


cnxzq (approved):

cnxzq (approved):

josh-hemphill:
The issue is that Argon 2 is not yet available in either Node core, or any large 3rd party modules. I think I brought it up when doing the rewrite. That's why I added the reference to the OWASP cheatsheets, since the status of Argon2 may change and the page not get updated to reflect the state of Argon2

goldbergyoni:
@tigranmaestro Welcome!

@josh-hemphill Which is the most reputable lib that implements Argon2?

josh-hemphill:
It looks like the Argon2 library has a lot more usage now, but it's a native binding, not pure JavaScript, so using it downloads additional binaries. Nodejs core is open to adding it, but Nodejs uses OpenSSL for all it's crypto operations, and OpenSSL won't add it until after OpenSSL 3.0 becomes stable. Another option would be to use the libsodium library which is large pure javascript crypto library which happens to include Argon2, but if that's all you need it for you'd want to make sure you're treeshaking.

goldbergyoni:
@josh-hemphill Great follow-up as always, sounds like waiting a bit more is the right act here?

josh-hemphill:
> @josh-hemphill Great follow-up as always, sounds like waiting a bit more is the right act here?

I personally think so. If we decide to mention Argon2 we're also going to have to explain how to choose between the native binaries and pure javascrypt implementations, and we'll have to update it again in a number of months. Currently, BCrypt and SCrypt cover all the bases for the *functionality* you'd need. I would also be worried about suggesting that devs use a library that may become irrelevant shortly, and as happens may stop being worked on shortly after.

goldbergyoni:
@josh-hemphill Resonates with me

goldbergyoni:
@josh-hemphill Can you close this PR based on your judgement?

josh-hemphill:
Yeah. Since Argon2 has been postponed until OpenSSL 3.1, it going to be a little while before we need to look at it again.

josh-hemphill:
Good news, Argon2 has been released in OpenSSL 3.2, and there's an open PR in Node to add it, so we should have a version of Node coming up that will include Argon2; can probably update the relevant info after that.

ranisalt:
> It looks like the Argon2 library has a lot more usage now, but it's a native binding, not pure JavaScript, so using it downloads additional binaries

There seems to be some misunderstanding here as [the library recommended for bcrypt](https://github.com/kelektiv/node.bcrypt.js) also contains native bindings, just like [node-argon2](https://github.com/ranisalt/node-argon2) (assuming it is the Argon2 library you mention). Both libraries have roughly the same install process, which downloads prebuilt binaries if available.

Is there something else that you see blocking?

PR #959, opened 2021-04-25
Add library-less example

### LINKED ISSUE #955: Expire JWT page refers to outdated / discontinued packages
https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/security/expirejwt.md

The last update to express-jwt-blacklist is 5 years ago, and the linked github on npm leads to a 404

https://www.npmjs.com/package/express-jwt-blacklist

josh-hemphill:
Closes #955 

Provides basic example of using a persistent key-value store (in this case Redis) to maintain a denylist to handle logouts. 

Removes the old example, as the library referenced is no longer maintained, and doing it with any key-value store is simple enough to not require any wrapping library.

### DIFF
--- a/sections/security/expirejwt.md
+++ b/sections/security/expirejwt.md
@@ -2,39 +2,43 @@
 
 ### One Paragraph Explainer
 
-By design, JWTs (JSON Web Tokens) are completely stateless, so once a valid token is signed by an issuer, the token may be verified as authentic by the application. The problem this leads to is the security concern where a leaked token could still be used and unable to be revoked, due to the signature remaining valid as long as the signature provided by the issues matches what the application is expecting.
-Due to this, when using JWT authentication, an application should manage a blacklist of expired or revoked tokens to retain user's security in the case a token needs to be revoked.
+By design, JWTs (JSON Web Tokens) are completely stateless, so once a valid token is signed by an issuer, the token may be verified as authentic by the application. The problem this leads to is the security concern where a leaked token could still be used and be unable to be revoked, due to the signature remaining valid as long as the signature provided by the issues matches what the application is expecting.
+Due to this, when using JWT authentication, an application should manage a denylist of expired or revoked tokens to retain user's security in the case a token needs to be revoked.
 
-### `express-jwt-blacklist` example
+### redis  example
 
-An example of running `express-jwt-blacklist` on a Node.js project using the `express-jwt`. Note that it is important to not use the default store settings (in-memory) cache of `express-jwt-blacklist`, but to use an external store such as Redis to revoke tokens across many Node.js processes.
+An example of using Nodejs and a Redis store containing your denylist. Keeping your denylist in an array or object is not recommended since if you restart, you'd lose it, and other instances wouldn't know if someone was explicitly logged out. So this assumes you have a running instance of Redis or some persistent key-value store and middleware handling login/JWT creation, and JWT verification since this is also not an extensive example of JWT implementation.
 
 ```javascript
-const jwt = require('express-jwt');
-const blacklist = require('express-jwt-blacklist');
+const redis = require("redis");
+const redisClient = redis.createClient();
 
-blacklist.configure({
-  tokenId: 'jti',
-  strict: true,
-  store: {
-    type: 'memcached',
-    host: '127.0.0.1',
-    port: 11211,
-    keyPrefix: 'mywebapp:',
-    options: {
-      timeout: 1000
+// Check for logout middleware
+app.use((req,res,next)=>{
+  const { token } = req;
+  redisClient.get(token, (error, data) => {
+    if (error) {
+      return res.status(400).send({ error });
     }
-  }
+    if (data !== null) {
+      return res.send({
+        message: 'You have to login!',
+      });
+    }
+    return next();
+  })
 });
 
-app.use(jwt({
-  secret: 'my-secret',
-  isRevoked: blacklist.isRevoked
-}));
-
 app.get('/logout', (req, res) => {
-  blacklist.revoke(req.user)
-  res.sendStatus(200);
+   const { userId, token, tokenExp } = req;
+      // Redis lets you set an expiration,
+      // so we'll use that to keep it clean of expired JWTs 
+      redisClient.set(token, tokenExp, true);
+
+      return res.send({
+        status: 'success',
+        message: 'Logout successful',
+      });
 });
 ```
 
@@ -42,3 +46,7 @@ app.get('/logout', (req, res) => {
 
 From the blog by [Marc Busqué](http://waiting-for-dev.github.io/blog/2017/01/25/jwt_secure_usage/):
 > ...add a revocation layer on top of JWT, even if it implies losing its stateless nature.
+
+### References
+
+See the [blog post by Tosin Moronfolu](https://dev.to/chukwutosin_/how-to-invalidate-a-jwt-using-a-blacklist-28dl) for a more detailed implementation guide.
--- a/sections/security/expirejwt.md
+++ b/sections/security/expirejwt.md
@@ -13,10 +13,14 @@ An example of using Nodejs and a Redis store containing your denylist. Keeping y
 const redis = require("redis");
 const redisClient = redis.createClient();
 
+// Make sure you handle the creation and verification of the JWT before checking its logout status
+
 // Check for logout middleware
 app.use((req,res,next)=>{
-  const { token } = req;
-  redisClient.get(token, (error, data) => {
+   const { userId, tokenIssuedAt } = req;
+
+  // Using the token itself as the matching key is not reliable, so use the content of the JWT
+  redisClient.get(userId + tokenIssuedAt, (error, data) => {
     if (error) {
       return res.status(400).send({ error });
     }
@@ -30,10 +34,10 @@ app.use((req,res,next)=>{
 });
 
 app.get('/logout', (req, res) => {
-   const { userId, token, tokenExp } = req;
+   const { userId, tokenExpiresAt, tokenIssuedAt } = req;
       // Redis lets you set an expiration,
       // so we'll use that to keep it clean of expired JWTs 
-      redisClient.set(token, tokenExp, true);
+      redisClient.set(userId + tokenIssuedAt, tokenExp, true);
 
       return res.send({
         status: 'success',


sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
-Due to this, when using JWT authentication, an application should manage a blacklist of expired or revoked tokens to retain user's security in the case a token needs to be revoked.
+By design, JWTs (JSON Web Tokens) are completely stateless, so once a valid token is signed by an issuer, the token may be verified as authentic by the application. The problem this leads to is the security concern where a leaked token could still be used and be unable to be revoked, due to the signature remaining valid as long as the signature provided by the issues matches what the application is expecting.
+Due to this, when using JWT authentication, an application should manage a denylist of expired or revoked tokens to retain user's security in the case a token needs to be revoked.
 
-### `express-jwt-blacklist` example
+### redis  example
```
rluvaton, line 8: 1. I believe there is an extra space between the _redis_ and the _example_
2. Isn't redis supposed to be Redis? 🤔

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
-    keyPrefix: 'mywebapp:',
-    options: {
-      timeout: 1000
+const redis = require("redis");
+const redisClient = redis.createClient();
+
+// Make sure you handle the creation and verification of the JWT before checking its logout status
+
+// Check for logout middleware
+app.use((req,res,next)=>{
```
rluvaton, line 19: I think you missed some spacing by mistake

rluvaton (changes requested):

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
+const redisClient = redis.createClient();
+
+// Make sure you handle the creation and verification of the JWT before checking its logout status
+
+// Check for logout middleware
+app.use((req,res,next) => {
+   const { userId, tokenIssuedAt } = req;
+
+  // Using the token itself as the matching key is not reliable, so use the content of the JWT
+  redisClient.get(userId + tokenIssuedAt, (error, data) => {
```
goldbergyoni, line 23: Shall we use a promisified version?

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
+    if (error) {
+      return res.status(400).send({ error });
```
goldbergyoni, line 25: Should this be 401 to signal auth issue? If not, then 500 seems more adequate

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
+    }
+    if (data !== null) {
```
goldbergyoni, line 27: HTTP status is more explicit (401, 403)?

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
-  isRevoked: blacklist.isRevoked
-}));
-
 app.get('/logout', (req, res) => {
-  blacklist.revoke(req.user)
-  res.sendStatus(200);
+   const { userId, tokenExpiresAt, tokenIssuedAt } = req;
+      // Redis lets you set an expiration,
+      // so we'll use that to keep it clean of expired JWTs 
+      redisClient.set(userId + tokenIssuedAt, tokenExp, true);
```
goldbergyoni, line 40: Is there a reason to store all the 50M users sessions in Redis and not only the blocked users?

goldbergyoni (changes requested):
Great start, clean code, many comments. Let's discuss and refine this a bit

goldbergyoni:
@animir I'm curious to hear your thoughts here

rluvaton:
@goldbergyoni We can add another reference from 


> **Why JWTs being stateless is awesome**
> JWTs are not meant to be stored in a database. In a distributed system, you might have several back-end services for different purposes and business domains. All these services need is a public key (more information on this below) and they can now verify tokens from incoming requests. There is no need to send a request to your auth server for every request (you have no idea how frequently I see this being done). This is a massive performance, resilience and scalability gain.
>
> From [JSON Web Tokens (JWT) — the only explanation you will ever need](https://link.medium.com/USRoQWL1igb)

animir:
@goldbergyoni @rluvaton Hi, first of all, it is good this PR is going to remove outdated `express-jwt-blacklist`.

I would pay attention to next:
1.  6.11 title says `Support blocklisting JWTs`, but context of the example in this PR is all about `logout`. It is quite the same, but we should be careful with words. In many other areas there are completely different meanings.
2. When JWT or (userId + something) is stored somewhere on store like Redis, why use it at all? Use cookie-based session or any other session approach. It is the same. My advice is to start from [this article on Auth0](https://auth0.com/docs/tokens/revoke-tokens)
3. This [blog post by Tosin Moronfolu](https://dev.to/chukwutosin_/how-to-invalidate-a-jwt-using-a-blacklist-28dl) doesn't look like a good reference. I have not checked code examples, may be they are good, may be not, but 22 likes for more than 2 months?

I hope, this helps.

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
+const redis = require("redis");
+const redisClient = redis.createClient();
+
+// Make sure you handle the creation and verification of the JWT before checking its logout status
+
+// Check for logout middleware
+app.use((req,res,next) => {
+   const { userId, tokenIssuedAt } = req;
+
+  // Using the token itself as the matching key is not reliable, so use the content of the JWT
```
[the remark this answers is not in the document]
josh-hemphill: From my comment on the issue: https://github.com/goldbergyoni/nodebestpractices/issues/955#issuecomment-821274849
Referencing the comment on the blog post: https://dev.to/hessman_/comment/1c9h9
Valid base64 encoded strings can have padding of "=" at the end. So one could maliciously log in with an expired token by just adding padding, since it would otherwise be a valid token.

sections/security/expirejwt.md
josh-hemphill: The redis library doesn't have native promise support. [NodeRedis/node-redis#promises](https://github.com/NodeRedis/node-redis#promises)
The readme tells users to manually promisfy the functions, so just to be concise in the example I used callbacks.

sections/security/expirejwt.md
josh-hemphill: I suppose I should phrase it better; it's keeping the redis store empty. The reason to use a denylist (over switching to stateful sessions or refresh tokens) with JWT is that the total number of denylisted tokens is extremely small, small enough that if you need to scale up, it would be practical to just have redis stores created with every node instance and actively push any new denylisted tokens to all your nodes.

sections/security/expirejwt.md
```diff
@@ -2,43 +2,55 @@
...
-    port: 11211,
-    keyPrefix: 'mywebapp:',
-    options: {
-      timeout: 1000
+const redis = require("redis");
+const redisClient = redis.createClient();
+
+// Make sure you handle the creation and verification of the JWT before checking its logout status
+
+// Check for logout middleware
```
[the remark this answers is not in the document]
josh-hemphill: Probably not, though it might make sense to mention both that it's checking against the denylist and that it covers logging out.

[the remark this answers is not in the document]
goldbergyoni: I believe that most developers will infer this alone, we can add few words on this. Throughout the code and explanations, I would stick everywhere to denylist

sections/security/expirejwt.md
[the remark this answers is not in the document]
goldbergyoni: Very interesting

sections/security/expirejwt.md
goldbergyoni: Should be 1 line long with with util.promisify. I suggest that we should not show anymore examples with callbacks.

goldbergyoni:
> @goldbergyoni @rluvaton Hi, first of all, it is good this PR is going to remove outdated `express-jwt-blacklist`.
> 
> I would pay attention to next:
> 
> 1. 6.11 title says `Support blocklisting JWTs`, but context of the example in this PR is all about `logout`. It is quite the same, but we should be careful with words. In many other areas there are completely different meanings.
> 2. When JWT or (userId + something) is stored somewhere on store like Redis, why use it at all? Use cookie-based session or any other session approach. It is the same. My advice is to start from [this article on Auth0](https://auth0.com/docs/tokens/revoke-tokens)
> 3. This [blog post by Tosin Moronfolu](https://dev.to/chukwutosin_/how-to-invalidate-a-jwt-using-a-blacklist-28dl) doesn't look like a good reference. I have not checked code examples, may be they are good, may be not, but 22 likes for more than 2 months?
> 
> I hope, this helps.

Thanks @animir 

1. Agree. cc @josh-hemphill 
2. This bullet doesn't advice to use JWT rather if you do use JWT for some reason (e.g. server to server integration), then allowing denylist is mandatory. We can few words on this "JWT does not suite all scenarios, if there are likely to be many denied or logged-out users then consider using a session-based approach"

sections/security/expirejwt.md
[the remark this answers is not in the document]
josh-hemphill: I wasn't sure if going into detail about that vulnerability would be extraneous. I'm happy to add more about it, since I suppose that might effect other things people would want to do with JWTs and would be good to know about regardless of whether they're implementing a denylist.

Yes that's it. Some suggest using the decoded signature as the matching key, but it's not text encoded as base64, so you'll probably need to convert it to some other format to store it. Though I just realized you could compare them as buffers using `Buffer.compare()`, so as long as you can reliably store the token as derived from binary/buffer format, then you could compare the whole tokens (though you'd need to remove the `.` separators for it to be valid base64). So perhaps the best bet is to just lop off the end segment `jwtString.split('.')[2]` and store and compare that using the base64 parsed to a buffer (which should take care of any padding.

josh-hemphill:
1. Yeah. I'll try to reword it when I get a chance, though I'm open to suggestions. I think we do need to make sure that the denylist is for both logging out users and terminating sessions.
2. I went into probably a little to much detail on a change comment: https://github.com/goldbergyoni/nodebestpractices/pull/959#discussion_r633197759 
Basically, in most systems, the number of logged out or terminated user sessions is extremely small, small enough that it's practical to just ad-hoc push any denylist updates to all your nodes. And because the tokens invalidate themselves after a period anyways, your denylist remains very small (only the subset that are still otherwise valid but need to be blocked)
The other solution that's often presented is **refresh tokens**, I'm open to switching the recommendation to that, but it has a higher upfront complexity (though doesn't need any special ad-hoc updates when at scale)
3. I'm open to switching to some other linked documentation, that was just the first google search result.

sections/security/expirejwt.md
animir: @josh-hemphill There is another popular Redis package [ioredis](https://www.npmjs.com/package/ioredis), it supports promises.

animir:
@josh-hemphill 
2. 
> the number of logged out or terminated user sessions is extremely small, small enough that it's practical to just ad-hoc push any denylist updates to all your nodes

That makes sense. As @goldbergyoni suggested, a note can be added there:

> We can few words on this "JWT does not suite all scenarios, if there are likely to be many denied or logged-out users then consider using a session-based approach"

3. From my understanding, reference should be from some widely-known person or popular article. I think, it is better to remove the link at all. Somebody can go later and put it, if there is suitable one.

sections/security/expirejwt.md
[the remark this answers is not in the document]
goldbergyoni: Very interesting

goldbergyoni:
@josh-hemphill LMK once I should review again. Take your time

josh-hemphill:
I've been doing some more thinking on this. I think there's actually two separate issues being discussed.
1. session-less authentication
2. JWT, the technology, that happens to come with a suggested protocol of session-less authentication, though you can use JWT without using the protocol as just a token generation algorithm

Because the more scalable solution is to use refresh tokens, which can still use JWT tokens with sessions, but use JWT as a convenient way of creating managed tokens that let you put the tokens where you want, e.g. in the request body (which can help keep tokens out of the logs of TLS terminating proxies) and manage multiple locally cached users in things like a kiosk environment, and decouples your authentication from reliance on cookies, though you could store them in cookies if you want.

I propose instead of making this a statement on JWT, perhaps it should just describe session-less authentication of any kind, that it still requires session-like expiration for things like logout and instant termination of a credential's access.

goldbergyoni:
@josh-hemphill Fully agree, would absolutely mention this, but the main message is still relevant.

Simply put, This item is about keeping a deny list in the right way + Remark that this is applicable mostly in session-less scenario (JWT as an example). 

p.s. Arent deny lists applicable to Sessions (e.g. Cookies) where one wishes to terminate attacker's access immediately?

josh-hemphill:
> p.s. Arent deny lists applicable to Sessions (e.g. Cookies) where one wishes to terminate attacker's access immediately?
@goldbergyoni , no, the point of having stateful sessions is that everything about the session is stored in the database already, the way to terminate their session would be to just delete their session from the database, or set their account to disabled if you need to block them entirely.
As it relates to refresh tokens, the point of using them is that the time between having to refresh your auth token is sufficiently short that only being able to terminate the session on the next token refresh is acceptable to you and/or that the reduction/elimination of session database lookups on every request leads you to needing to implement some acceptable latency. Unlike session-less tokens, token refreshes happen in the background without user interaction, so they can be set to just some number of minutes to almost eliminate any potential risk that the session termination latency introduces.

goldbergyoni:
@animir Side question - I'd like to introduce you to some rate limiting challenge and hear your thoughts (probably will be implemented using node-rate-limiter-flexible). Can you PM via me@goldbergyoni.com?

goldbergyoni:
@josh-hemphill Can you close this PR based on your judgment?

josh-hemphill:
I completely forgot I was working on this, sorry.
I'm not sure, I still think we need to revise the token invalidation stuff. I'm thinking blacklists might actually be more technically complicated than refresh tokens, and are really the domain of architecting distributed systems, which is why most of the documentation about it is for JWT API keys.

PR #966, opened 2021-04-30
Convert to github actions

rluvaton:
As discussed (https://github.com/goldbergyoni/nodebestpractices/issues/896#issuecomment-826141118), I've migrated from Travis.CI to GitHub Actions

I've sinned, I've also Fixed that the tests would run in this repo as well and not only in the forked ones (https://github.com/goldbergyoni/nodebestpractices/issues/896#issuecomment-810954302)

- [x] Remove `.travis.yml` file


.github/workflows/gen-html.yml
```diff
@@ -0,0 +1,29 @@
...
+      uses: actions/setup-node@v2.1.2
+      with:
+        node-version: '14'
+
+    - run: cd .operations && npm install
+    - run: cd .operations && npm run lint
+    - run: cd .operations && npm run build
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        IS_FORK: ${{ github.event.pull_request.head.repo.full_name != 'goldbergyoni/nodebestpractices' }}
```
rluvaton, line 29: haven't checked but this may cause problems if we run on a push event

rluvaton: Fixed

elektronik2k5 (approved):

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Setup Node.js environment
+      uses: actions/setup-node@v2.1.2
+      with:
+        node-version: '14'
+
+    - run: cd .operations && npm install
```
goldbergyoni, line 24: Consider:

```
defaults:
  run:
    shell: bash
    working-directory: .operations
    ```

File: .operations/gen-html.js [hunk 3]
```diff
@@ -7,6 +7,7 @@ const { readdir, readFile, writeFile } = require('graceful-fs');
 const imagemin = require('imagemin');
```
goldbergyoni, line 7: @rluvaton Can you kindly remind me what is this file about? Is this new one or an existing file?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+    - run: cd .operations && npm run lint
```
goldbergyoni, line 25: Should there be a linting issue, will we see it within the code in the GitHub "files" tab?

goldbergyoni (changes requested):
@rluvaton Great stuff here. Don't have any major concern, mostly asked few questions to refresh my memory

.github/workflows/ci.yml
rluvaton: I don't think so, but I can check now by introducing a linting issue

.operations/gen-html.js
rluvaton: This is an existing file and I think it's about generating GitHub Pages, but I'm not sure 🤔

goldbergyoni: @BrunoScheufler @js-kyle Do you remember what is the purpose of this file? Is it needed anymore?

.github/workflows/ci.yml
goldbergyoni: It's possible, should work by default given the default eslint behaviour

This extension enriches the DX:

https://github.com/marketplace/actions/lint-action

rluvaton: When we add eslint we should the GitHub action you suggested, but right now we're using `markdownlint` which doesn't support it (I haven't found a GitHub Action that support that, but I can keep looking)

goldbergyoni (approved):
LGTM

.operations/gen-html.js
rluvaton: Do you want to wait for them to answer or can we merge now?

goldbergyoni:
@rluvaton Feel free to merge

PR #1028, opened 2021-07-04
Fix update last update badge no escape when replacing img tag

rluvaton:
Fixed bug caused by me 😬 in #1027 

The error was:
```
sed: -e expression #1, char 244: unknown option to `s'
```

~Haven't tested it yet, cause I did a quick fix on my phone...~

Tested it and it worked.


elektronik2k5 (approved):

rluvaton:
Thanks @elektronik2k5

PR #1033, opened 2021-07-11
SeparateExpress: export app!

SConaway:
Noticed that it was missing, which could be confusing since otherwise the `www` file is creating a new instance of a nothing.

### DIFF
--- a/sections/projectstructre/separateexpress.md
+++ b/sections/projectstructre/separateexpress.md
@@ -15,6 +15,8 @@ const app = express();
 app.use(bodyParser.json());
 app.use('/api/events', events.API);
 app.use('/api/forms', forms);
+
+module.exports = app;
 ```
 
 ### Code example: Server network declaration, should reside in /bin/www


goldbergyoni:
@SConaway Welcome aboard and thanks for brining this interesting discussion 🌻

expess-generator is just one way of structuring an app and I'm not sure that I would recommend it:

1. HTTP is just entry-point to the app -  Going with the generator philosophy, WWW is the starting point of the code. In many backends, there are other doors like message queues, scheduled jobs so the right structure is: start.js calling all the entry-points to start listening. Express is just one of them

2. It's less testable - If the first thing that the app does is open network sockets, the integration tests get no chance to configure stuff beforehand. Ideally, the express/API has a 'start' method which the test can call +. 'stop' method that close connection

Thoughts?

SConaway:
Yes, I agree. However, I'm not sure how exporting app would go against this structure. if you generate a new app like that, you'll find that app is exported by default.

goldbergyoni:
@SConaway See here an example app:

https://github.com/testjavascript/nodejs-integration-tests-best-practices/blob/master/example-application/entry-points/api.js

PR #1094, opened 2021-11-08
removed typo

Tosin-webdev:
This removes the comma.


elektronik2k5 (changes requested):
So the comma removal is fine, but the code style changes and other markdown changes look like automatic opinionated reformatting and unintended.

Tosin-webdev:
Alright, thanks for the feedback.

PR #1124, opened 2021-12-02
fix lint issues

rluvaton:


rluvaton:
Still, there is one lint error which needed to be fixed

elektronik2k5 (approved):

PR #1287, opened 2023-11-11
Update limitrequests.md

animir:
Use ioredis package

### DIFF
--- a/sections/security/limitrequests.md
+++ b/sections/security/limitrequests.md
@@ -8,12 +8,10 @@ Rate limiting should be implemented in your application to protect a Node.js app
  
   ```javascript
  const http = require('http');
- const redis = require('redis');
+ const IoRedis = require('ioredis');
  const { RateLimiterRedis } = require('rate-limiter-flexible');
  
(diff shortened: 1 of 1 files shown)


animir:
@goldbergyoni Above lint error isn't related to this PR.
We should update to `ioredis` package in the example since previously used `redis` package has changed how they create Redis client. It is more complicated now and not the best for simple example like that. Current example in `master` branch doesn't work with the latest `redis` package. With `ioredis` it works for any version.

goldbergyoni:
@animir Great point here, thanks!

And... congrats for the new release of node-rate-limiter-flexible 🥳

Let's make some article or YouTube about rate limiting some day

animir:
@goldbergyoni Hi, thanks!
There are some articles about rate limiting and brute-force protection on Medium I wrote a couple of years ago https://medium.com/@animirr
I see you do have YouTube channel. It may be interesting to collaborate.
I'll make sure to email you to continue this discussion.

Fool256 (approved):

Fool256 (approved):

